PR #10, opened 2024-09-24
parse vllm csv report

coketaste:
parse vllm csv report
extract performance from throughput and latency


gargrahul:
Closing in favor of #14

PR #18, opened 2024-11-27
Refine GPU check

gargrahul:
Fix get_gpu_vendor function as it only checks for installation of smi functions, not checking if it is correctly working.


tools/utils.py
```diff
@@ -541,6 +541,13 @@ def generate_report(
 # Utility functions
 # ==================================================================================================
 
+from typing import List
+
+def subprocess_run(cmd: List[str]):
```
amathews-amd, line 546: standardize on console.sh

PR #50, opened 2025-05-21
[vllm online serving] adding ignore-eos

gargrahul:
Adding
--ignore-eos
at vllm online serving to give consistent performance across different runs. Without this option, the requests could terminate earlier and perf will shift randomly in random datasets.


Rohan138 (approved):
LGTM

PR #55, opened 2025-05-30
Update path to use profiling scripts in VLLM

gargrahul:


scripts/vllm/vllm_benchmark_report.sh
```diff
@@ -87,8 +87,8 @@ tag="vllm_rocm6.3.1"
 
 report_dir="reports_${datatype}_${tag}"
 report_summary_dir="${report_dir}/summary"
-tool_latency="/app/vllm/benchmarks/benchmark_latency.py"
-tool_throughput="/app/vllm/benchmarks/benchmark_throughput.py"
+tool_latency="/app/vllm/benchmarks/profiling/benchmark_latency.py"
+tool_throughput="/app/vllm/benchmarks/profiling/benchmark_throughput.py"
```
ppalaniappan-amd, line 91: I am under the impression ROCm/vLLM will be deprecated and vLLM would be built from upstream repo - https://github.com/vllm-project/vllm/tree/main/benchmarks
However, the "profiling" folder is only available in ROCm/vLLM - https://github.com/ROCm/vllm/tree/main/benchmarks/profiling
Please check if this can lead to potential issues.

Rohan138: Yes, this is a known issue-we're moving to the profiling script to maintain perf for this release because the benchmarking script has changed default configs in upstream.
For the future releases we're talking to the vLLM team; either they'll file an upstream PR to keep the legacy benchmark config available, or we may have to move these scripts into MAD.

PR #56, opened 2025-05-30
vLLM 05/27 release

gargrahul:


benchmark/vllm/README.md
```diff
@@ -317,11 +317,6 @@ owners and are only mentioned for informative purposes.   
 ----------
 This release note summarizes notable changes since the previous docker release.
 
--   The vLLM version number was incremented from 0.8.3 to 0.8.5.
```
ppalaniappan-amd, line 320: Should we save history of changes, rather than just comparing against previous release?

gargrahul: Commit history can be used for this ?

ppalaniappan-amd (approved):
LGTM

PR #62, opened 2025-06-30
Parse and load docker build args

coketaste:
Add logic to parse and load docker build args

### DIFF
--- a/tools/run_models.py
+++ b/tools/run_models.py
@@ -255,7 +255,18 @@ def run_model(
     if args.clean_docker_cache:
         use_cache_str = "--no-cache"
 
-    build_args = ""
+    # Set the Docker build arguments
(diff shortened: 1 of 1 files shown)


tools/run_models.py
```diff
@@ -255,7 +255,18 @@ def run_model(
     if args.clean_docker_cache:
         use_cache_str = "--no-cache"
 
-    build_args = ""
+    # Set the Docker build arguments
+    docker_build_args = {"MAD_SYSTEM_GPU_ARCHITECTURE": get_system_gpu_arch()}
```
shikamd123, line 259: can we add an error check if the variable might have an empty value , it should fail right away and show the error message ?

coketaste: The function get_system_gpu_arch() already includes an error handler.

shikamd123 (approved):

shikamd123:
Looks good to me.

PR #76, opened 2025-08-13
VLLM refactor and 08/12 update

gargrahul:
## Technical Details

-     Added default, extended, and performance configs
-     Removed deprecated models (Llama 2 7B, Mistral 7B, Qwen 2 7B, Qwen 2 72B, Gemma 2 27B, DeepSeek 16B MoE, DBRX Instruct, Falcon 180B)
-     Updated scripts to use configs and allow overrides using madengine CLI
-     Soft-deprecate latency benchmark in favor of online serving
-     Enable AITER FP8 KV Cache


Rohan138 (approved):
LGTM

PR #83, opened 2025-09-09
Added Qwen3-30B-A3B to MAD, using single GPU

gargrahul:


Rohan138 (dismissed):
Needs one change, AITER_MHA is still incorrectly pinned

lonchchu (approved):

PR #90, opened 2025-09-17
JAX-Maxtext training v25.7 JAX 0.6.0 patch

gargrahul:
Patch for Jax-Maxtext v25.7 using Jax version 0.6.0
The following update

- Fixes nanoo FP8 issue in v25.7 Jax 0.6.0
- Improves performance of select models


yeandy:
LGTM

PR #126, opened 2026-01-22
Remove MPT-30B until huggingface model link is back

EricKing626:
Since MPT-30B is removed from Huggingface, the MAD repo should temporarily remove this model.


lonchchu (approved):
The change looks good. Thanks!

PR #179, opened 2026-07-22
Add new vLLM models e.g. Kimi K2, DSV4, GLM

gargrahul:


Rohan138 (approved):

leconcio:
everything else looked good, just want to know if we meant to remove both llama models

Rohan138:
@leconcio yes-we're deprioritizing the Llama 4 models from the vLLM regression tests run through MAD. They'll still have limited UT coverage in vllm CI.

Cemberk (approved):
LGTM 👍

PR #200, opened 2026-08-10
sglang-disagg: fix queue-pair creation on Broadcom Thor2 (rdma-core 63.0), and make every overlay stage optional

mkuznet1:
`aicomnet-305/sgl-dev-gfx950-rdma63` -> `mad-rccl`. One commit, one file:
`docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile`.

## The problem

None of the KV-transfer backends could run 2P/2D DeepSeek-R1 on the Broadcom Thor2
(`bnxt_re`) fabric, and each failed in its own way, which made it look like three
unrelated bugs:

- mooncake logged 2604 `Failed to create QP` over a single run
- NIXL never started: UCX creates one interface queue pair while opening
  `rc_verbs/bnxt_re1`, and treats the failure as fatal
- MoRI built queue pairs that reached RTS, then moved zero bytes and aborted every
  request

All three are the same defect. The `bnxt_re` provider in rdma-core 39.0/50.0 corrupts
the verbs command buffer for work issued off worker threads, so the kernel answers
`ibv_create_qp` with EFAULT (errno 14). Isolated on one node, in one container, on the
same kernel, with 8 threads x 64 queue pairs and only the userspace changing:

| rdma-core | queue pairs created |
|-----------|---------------------|
| 39.0 (distro) | 33-66 of 512 |
| 63.0 (source) | 512 of 512 |

## What changed

**rdma-core.** The stage that fixes this was already in this file as `ENABLE_RDMA62`.
It worked, but it was off by default and its comment said only "hosts whose RDMA stack
needs a newer libibverbs", so nothing connected it to Thor2 and it was never switched
on. It is now `RDMA_CORE_VERSION` (63.0, on by default; empty keeps the base image's),
documented against the measurement above, and it additionally:

- removes the distro rdma-core packages, so one libibverbs is left on disk instead of
  two of different vintages;
- drops the vendor `bnxt_re` provider that `/etc/ld.so.conf.d` puts on the loader path.
  It advertises kernel uABI 7-8, so on a host running the upstream driver (uABI 1) every
  device is rejected with `Driver bnxt_re does not support the kernel ABI of 1`, after
  which RCCL finds no IB plugin at all.

Ordering is load-bearing and is called out in the file: the source build replaces the
distro packages with `dpkg -r --force-all`, which leaves still-installed dependents
(libucx0, openmpi, ...) with dangling deps, so every apt operation has to run before the
removal. That is why the stage is last.

**Base image and stage gating.** The default base becomes
`rocm/sgl-dev:v0.5.16-rocm720-mi35x-20260807`, which already carries sglang, aiter,
sgl-kernel, MoRI, RIXL and mooncake built for gfx950. Rebuilding those on top of it is
both slow and a regression risk, so each component stage is now gated and defaults to
off: the image is the base plus rdma-core, and each `ENABLE_*` arg turns one component
into an A/B against what the base ships.

**Mooncake.** Moves from the `0.3.6.post1` wheel to a source build of `v0.3.12.post1`
with `ENABLE_MULTI_PROTOCOL=ON`. Before upstream 45b84d3 (#2725) there is no cross-host
guard, so a GPU address registered under both rdma and hip keeps hip's priority for any
target: every cross-node KV chunk is pushed through the intra-node-only HIP transport
and dies in `hipIpcOpenMemHandle`. Because the guard sits behind a CMake option that
defaults to OFF, a build without it produces the same broken routing silently, so the
stage greps the installed package for `MC_DISABLE_HIP` and fails the build if the guard
did not land.

## Compatibility

Existing users of this file get a different default image, so this is a behaviour
change. The old configuration is still reachable in full:

```
--build-arg BASE_DOCKER=lmsysorg/sglang:v0.5.12.post1-rocm720-mi30x \
--build-arg BUILD_GPU_TARGETS=gfx942 --build-arg MORI_GPU_ARCHS=gfx942 \
--build-arg ENABLE_RCCL_OVERLAY=1 --build-arg ENABLE_MORI_OVERLAY=1 \
--build-arg ENABLE_NIXL_OVERLAY=1 --build-arg ENABLE_MOONCAKE_OVERLAY=1
```

Hosts that do not want the new rdma-core can pass `--build-arg RDMA_CORE_VERSION=` to
keep the base image's.

### DIFF
--- a/docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
+++ b/docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
@@ -12,14 +12,20 @@
 # sgl-kernel, MoRI, RIXL and Mooncake built for the right GPU target, so every
 # component stage below defaults to OFF and the image is the base plus rdma-core.
 # Each stage is a substitution knob: turn one on to A/B that component against
-# what the base ships. Pick the base tag whose suffix matches the GPUs
-# (`-mi35x` = gfx950/MI35x, `-mi30x` = gfx942/MI30x) and set BUILD_GPU_TARGETS
-# and MORI_GPU_ARCHS to match.
+# what the base ships.
+#
+# The target GPU is chosen in exactly two places: BASE_DOCKER, whose tag suffix
+# names the GPUs (`-mi35x` = gfx950/MI35x, `-mi30x` = gfx942/MI30x), and
+# BUILD_GPU_TARGETS, which every stage that compiles device code derives from
+# (MORI_GPU_ARCHS defaults to it). Set those two consistently and nothing else
+# needs touching.
 #
 # Stages, all gated by build args (override via --build-arg):
 #   rdma-core : RDMA_CORE_VERSION (default 63.0, the only stage ON by default)
 #               — see the stage comment; this is what makes queue-pair creation
-#               work on Broadcom Thor2 (bnxt_re).
+#               work on Broadcom Thor2 (bnxt_re). Pass an empty value
+#               (RDMA_CORE_VERSION=) to keep the base image's rdma-core: the fix
+#               is opt-out, not mandatory.
 #   RCCL      : ENABLE_RCCL_OVERLAY=1 → ROCm/rocm-systems develop @ RCCL_COMMIT
 #               (default 78e8ba0) + smifix
 #   MoRI      : ENABLE_MORI_OVERLAY=1 → ROCm/mori @ MORI_COMMIT
@@ -36,8 +42,9 @@
 #
 # To reproduce the pre-sgl-dev behaviour (build everything on a plain sglang
 # base), pass BASE_DOCKER=lmsysorg/sglang:v0.5.12.post1-rocm720-mi30x
-# BUILD_GPU_TARGETS=gfx942 MORI_GPU_ARCHS=gfx942 and set the four ENABLE_* args
-# to 1.
+# BUILD_GPU_TARGETS=gfx942, set the four ENABLE_* args to 1, and pass an empty
+# RDMA_CORE_VERSION= — without that last one the old recipe still replaces the
+# base image's rdma-core, which the old file never did.
 ###############################################################################
 ARG BASE_DOCKER=rocm/sgl-dev:v0.5.16-rocm720-mi35x-20260807
 FROM $BASE_DOCKER
@@ -48,7 +55,7 @@ USER root
 ###############################################################################
 # 1) RCCL overlay — rebuild RCCL from source so the RCCL under test wins over
 #    the base image's librccl. (mirrors sglang_disagg_inference_rccl_overlay)
-#    OFF by default: the sgl-dev base already carries an librccl with the right
+#    OFF by default: the sgl-dev base already carries a librccl with the right
 #    code objects, and keeping it makes the base the control arm of the A/B.
 ###############################################################################
 ARG ENABLE_RCCL_OVERLAY=0
@@ -151,7 +158,9 @@ ARG MORI_REPO=https://github.com/ROCm/mori.git
 ARG MORI_BRANCH=main
 ARG MORI_COMMIT=a14e6992ffa95478e83127fe2672afff2840856f
 ARG MORI_WHEEL_URL=
-ARG MORI_GPU_ARCHS=gfx950
+# Derived, so the GPU target is set once via BUILD_GPU_TARGETS. Override only to
+# build MoRI for a different arch than the rest of the image.
+ARG MORI_GPU_ARCHS=${BUILD_GPU_TARGETS}
 ARG MORI_VERSION=1.2.0
 ARG MORI_SRC_DIR=/sgl-workspace/mori
 
@@ -179,6 +188,7 @@ RUN set -e; \
       pip install --no-cache-dir --force-reinstall "${MORI_WHEEL_URL}"; \
     else \
       echo "[mori-overlay] source build ${MORI_REPO}@${MORI_BRANCH}${MORI_COMMIT:+ (${MORI_COMMIT})} archs=${MORI_GPU_ARCHS}"; \
+      sed -i 's|http://|https://|g' /etc/apt/sources.list 2>/dev/null || true; \
       apt-get -o Acquire::Retries=5 update; \
       DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
         git cmake ninja-build pkg-config make patch; \
@@ -338,6 +348,7 @@ ARG MOONCAKE_HIP_DMABUF=ON
 
 RUN if [[ "${ENABLE_MOONCAKE_OVERLAY}" == "1" ]]; then \
       set -e; \
+      sed -i 's|http://|https://|g' /etc/apt/sources.list 2>/dev/null || true; \
       apt-get -o Acquire::Retries=5 update; \
       DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
         git cmake ninja-build build-essential pkg-config patchelf; \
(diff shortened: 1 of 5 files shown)


docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
```diff
@@ -1,28 +1,45 @@
...
+#               (default v0.3.12.post1, the first release with the cross-host
+#               guard)
+#
+# To reproduce the pre-sgl-dev behaviour (build everything on a plain sglang
+# base), pass BASE_DOCKER=lmsysorg/sglang:v0.5.12.post1-rocm720-mi30x
+# BUILD_GPU_TARGETS=gfx942 MORI_GPU_ARCHS=gfx942 and set the four ENABLE_* args
+# to 1.
 ###############################################################################
-ARG BASE_DOCKER=lmsysorg/sglang:v0.5.12.post1-rocm720-mi30x
+ARG BASE_DOCKER=rocm/sgl-dev:v0.5.16-rocm720-mi35x-20260807
```
i-kosarev, line 49: The default base plus BUILD_GPU_TARGETS/MORI_GPU_ARCHS now target gfx950, but sglang-disagg-deepseek-r1-overlay in models.json has no skip_gpu_arch and MAD only passes MAD_SYSTEM_GPU_ARCHITECTURE, which this Dockerfile never reads. I am afraid a normal MAD run on a gfx942 host will therefore silently build and run the MI35x image — please either restrict the model card to gfx950 or derive BASE_DOCKER/BUILD_GPU_TARGETS from the detected architecture.

docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
```diff
@@ -236,54 +310,162 @@ RUN set -e; \
...
+# distro libibverbs/librdmacm packages via `dpkg -r --force-all`, which leaves
+# still-installed dependents (libucx0, openmpi, ...) with dangling deps, so any
+# later apt command aborts with "Unmet dependencies". Every apt operation must
+# run before the dpkg removal; only dpkg and `ninja install` may follow it. This
+# is also why the stage is last.
 ###############################################################################
-ARG ENABLE_RDMA62=
-ARG RDMA_VER=v62.0
-RUN if [[ -n "${ENABLE_RDMA62}" ]]; then \
+ARG RDMA_CORE_VERSION=63.0
```
i-kosarev, line 398: scripts/sglang_disagg/README.MD:98 still documents MOONCAKE_VERSION and ENABLE_RDMA62 for this Dockerfile, but both are gone — renamed to MOONCAKE_REF and RDMA_CORE_VERSION, the latter now ON by default rather than opt-in. Docker silently ignores unknown build args, so anyone following the README today gets a different image than they expect.

docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
```diff
@@ -236,54 +310,162 @@ RUN set -e; \
...
-# Sanity: mooncake must import alongside nixl/rixl (non-fatal — module path may vary).
-RUN python3 -c "import mooncake; print('MOONCAKE_OVERLAY_OK')" \
-    || echo "MOONCAKE_IMPORT_DIFFERS (verify sglang mooncake import path at runtime)"
+# Report which mooncake is in place and whether it can route a cross-node
+# transfer around the HIP transport at all. Informational only: never fail the
+# build here (see the RCCL report above for why).
+RUN ( set +e +o pipefail; \
+      pkg="$(python3 -c 'import mooncake, os; print(os.path.dirname(mooncake.__file__))' 2>/dev/null)"; \
+      echo "MOONCAKE_PACKAGE ${pkg:-<not found>}"; \
+      echo "MOONCAKE_CROSS_HOST_GUARD ${pkg:+$(grep -rls MC_DISABLE_HIP "$pkg" | tr '\n' ' ')}" )
```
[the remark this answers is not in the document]
mkuznet1: Not reproducible. The nested quotes are inside `$( )`, which starts a fresh quoting context, so bash parses the line fine. This layer built three times in validation (jobs 25000/25004/25008) and the line printed `MOONCAKE_CROSS_HOST_GUARD /opt/venv/lib/python3.10/site-packages/mooncake/store.cpython-310-x86_64-linux-gnu.so ...`. Closing.

docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
```diff
@@ -31,82 +48,110 @@ USER root
 ###############################################################################
 # 1) RCCL overlay — rebuild RCCL from source so the RCCL under test wins over
 #    the base image's librccl. (mirrors sglang_disagg_inference_rccl_overlay)
+#    OFF by default: the sgl-dev base already carries an librccl with the right
```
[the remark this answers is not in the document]
mkuznet1: Fixed in d725fa9.

docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
```diff
@@ -127,11 +172,17 @@ WORKDIR /sgl-workspace
...
 RUN set -e; \
-    if [[ -n "${MORI_WHEEL_URL}" ]]; then \
+    if [[ "${ENABLE_MORI_OVERLAY}" != "1" ]]; then \
+      echo "[full-overlay] MoRI overlay disabled — keeping the image's MoRI"; \
+    elif [[ -n "${MORI_WHEEL_URL}" ]]; then \
       echo "[mori-overlay] installing prebuilt wheel: ${MORI_WHEEL_URL}"; \
       pip install --no-cache-dir --force-reinstall "${MORI_WHEEL_URL}"; \
     else \
       echo "[mori-overlay] source build ${MORI_REPO}@${MORI_BRANCH}${MORI_COMMIT:+ (${MORI_COMMIT})} archs=${MORI_GPU_ARCHS}"; \
+      apt-get -o Acquire::Retries=5 update; \
```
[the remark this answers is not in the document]
mkuznet1: Added in d725fa9 to all three stages (MoRI, mooncake, rdma-core). Line 113 needs nothing: it sits inside the same ENABLE_RCCL_OVERLAY block where line 72 has already rewritten sources.list. Robustness only — the default sgl-dev base serves every repo over https, so no build was failing on this.

docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
mkuznet1: Real gap, but it predates this PR: the card had no skip_gpu_arch before either, and the old default base was -mi30x/gfx942, so a gfx950 host silently got the wrong image too. Proper arch-aware selection is tracked separately. d725fa9 narrows it: the GPU target is now set only via BASE_DOCKER + BUILD_GPU_TARGETS, with MORI_GPU_ARCHS derived from the latter.

docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile
mkuznet1: Fixed in d725fa9: the table now lists RDMA_CORE_VERSION (default 63.0, empty value to opt out), MOONCAKE_REF and the ENABLE_* gates, and the text states the component stages are off by default. The same two dead args were also in the mad-slurm-multinode skill references and its manifest template, updated there too.

mkuznet1:
d725fa9 also covers the suppressed review notes: the legacy-reproduction recipe now carries RDMA_CORE_VERSION= (without it that recipe still replaced the base image's rdma-core), and the stage-5 comment is back to six lines.

i-kosarev (approved):

PR #176, opened 2026-07-08
vllm_dissag: GLM-5.1/5.2-FP8 (MLA+DSA) MoRI-EP WideEP disaggregated enablement

raviguptaamd:
## Summary
Adds **GLM-5.1-FP8** (`GlmMoeDsaForCausalLM` = MLA + DeepSeek Sparse Attention) to the MoRI-EP WideEP disaggregated serving path, stacked on the #171 unified launcher. Fully isolated from DeepSeek-V3/R1 — GLM gets its own image and a `MODEL_NAME`-gated runtime path, so existing models are byte-identical to `develop`.

## Defects fixed
Validated on **1P/1D (EP8)** and **2P/2D (EP16)**: NIAH 2k–35k = **10/10, no crash**; DeepSeek-V3 non-regressed.

- **Long-context accuracy collapse** (garbage >~30k tokens): vLLM #47766 cache-key fix keeps the persistent sparse-MLA kernel **ON** (keys attention metadata on per-request context+query length instead of a colliding key under chunked prefill).
- **8k disagg prefill crash**: DSA adds a *second* (indexer) KV cache per layer that the single-geometry MoRIIO connector never transferred → decode `unmap MISS` → 60s deferred-write expiry → prefill EngineCore crash. Now the indexer cache is paired with its main cache and shipped prefill→decode. Plus the DSA invalid-token kernel fix (#45324) and the shik-latest DP-notify/hash-failsafe path.

## Changes
- **`docker/vllm_disagg_inference.glmv5.1.ubuntu.amd.Dockerfile`** — NEW per-model image (GLM vLLM + aiter `e03fa6040` + mori `42e895472b08` + router). The base `vllm_disagg_inference` Dockerfile (DSV3/R1) is **left untouched**. Future models (e.g. Kimi-2.6) add their own Dockerfile the same way.
- **`models.json`** — card `pyt_vllm_disagg_mori_glm-5.1-fp8` → the new Dockerfile (`GLM_SKIP_PATCHERS=1`: the image carries the DSA fixes in-source).
- **`models.yaml`** — GLM-5.1-FP8 recipe (block=1, AITER MLA on, eager, per-role mori backends). Also DeepSeek-V3 `dp:` caps (`--max-num-seqs 64 --max-model-len 32768`) to bound the newer base's decode split-KV logits workspace — **isolated to the DSV3 entry**.
[...]

PR #205, opened 2026-08-17
vllm_disagg: GLM-5.2 (FP8 + MXFP4) on MI355X/gfx950 + AI NIC, 1P/1D EP8 [stacked on #176]

raviguptaamd:
> **Overlap with #176 — please read first.** This branch is stacked on #176
> (`glm5.1-wideep-clean`), which is still open. `develop` is the only long-lived branch in
> this repo, so GitHub cannot express the stack: this diff therefore contains **#176's 8
> commits as well as this work's 8**. #176 is the GLM-5.**1** wideEP enablement and has its
> own review; **only the 8 commits below are new here.**
>
> ```
> bef784f  document the GLM-5.2 MI355X recipe
> 801e536  parse_to_csv: stop dropping the first benchmark cell; fix backend and arch tags
> adc163c  niah: make the accuracy check a real pass/fail gate, jitter needle placement
> 24b5c9c  make the launcher portable across sites, and add the AAC/ionic env file
> d40c5e3  models.yaml: add GLM-5.2-FP8 and GLM-5.2-MXFP4 recipes for MI355X
> 500353f  moriio: pass RDMA tuning via extra_config, widen GLM gate, disable 2 harmful patchers
> 0f660fe  build the GLM image for gfx950 (MI355X) + ionic, and fix a silent wrong-arch bug
> 69eb062  remove stray NFS silly-rename artifact
> ```
>
> Reviewing just those: `git diff 550603f..HEAD`, or
> https://github.com/raviguptaamd/MAD/pull/3 which is the same 8 commits scoped to 11 files.
> Merge #176 first and this collapses to that clean diff.

Enables **GLM-5.2 (FP8 + MXFP4) on MI355X (gfx950) with AMD AI NIC (ionic)** for 1P/1D EP8
disaggregated serving over MoRI-IO. Measured on AAC, 2 nodes x 8 x MI355X, ROCm 7.2.3.

## Result

**TPOT target met.** ISL/OSL 28,672/1,024, FP8, EP8, 1P/1D:

| concurrency | 16 | 32 | 64 |
|---|---|---|---|
| TPOT (ms) | 33.7 | 36.4 | 40.2 |

Target is 50 ms avg. Before this change it was 104 ms.

**TTFT target not met, and this PR does not fix it.** Prefill measures 6,135–8,359
tok/s/rank against a 34,000 target (~4–5x short), and the fitted single-request TTFT at
80K ISL is ~13.7 s against a <7 s SLO. `GLM52_MI355X.md` §5 records the profile behind that
(MLA sparse attention ~82% of prefill, all GEMM 17.8%, MoE 2.9%), the six levers already
closed by measurement so they are not re-litigated, and why PCP is not the near-term answer.

## The two load-bearing values

**`DECODE_CUDAGRAPH_MODE: FULL_AND_PIECEWISE`** — the single change that moved TPOT
104 ms -> 34–40 ms. `PIECEWISE` splits the decode graph on the 3 DSA ops x 78 layers,
~234 launch boundaries per step. That cost is *batch-independent*, so it presents as a
latency floor that does not move with concurrency, batch size, or fabric — which is why it
survived so much tuning before being found. Co-requisite: `use_inductor_graph_partition`
must stay ON and no bare `--enforce-eager`, or boot and warmup both pass and the first real
MLA decode dies.

**`prefill.dp: --gpu-memory-utilization 0.72`** — the MLA chunked-prefill workspace is sized
from a hardcoded 64k-token clamp (`determine_chunked_prefill_workspace_size`), not from
`max_num_batched_tokens`, so it allocates 3.50 GiB on top of the KV pool regardless of
scheduler batch size — lazily, on the first long prefill, i.e. after boot and warmup have
both reported healthy. 0.80 OOM'd.

## A silent wrong-arch bug likely affecting other Dockerfiles here

The base image `rocm/vllm-dev:ci_base-*` exports `PYTORCH_ROCM_ARCH` and `MAX_JOBS` as
**ENV**, which shadows the same-named `ARG` in the child Dockerfile. So
`ARG PYTORCH_ROCM_ARCH="gfx950"` was silently inert and the build produced gfx942 kernels
while appearing to succeed. Fixed by renaming to `BUILD_ROCM_ARCH` / `BUILD_MAX_JOBS` and
assigning through. Caught only because the base's `MAX_JOBS` happened to be empty
(`int('')` crash) — otherwise it fails silently, so it is worth checking other child
Dockerfiles in this tree for the same shadowing.

## Also in here

- **`connectors/moriio.env.aac`** — AAC/ionic platform env: 8 rails
  `rocep{9,25,105,121,137,153,233,249}s0`, `MORI_IB_GID_INDEX=1`, mgmt NIC for bootstrap.
- **Launcher portability** — `FABRIC_SUBNET` defaulted to the MI300X cluster's `10.158.`
  prefix; on AAC (`10.2.80.`) that advertises a `MASTER_ADDR` the peer cannot reach and the
  run hangs at `Waiting for nodes` forever with no error. Unlike every other key in the env
  file this one is consumed **on the host**, so the `.env` loader needed an explicit one-key
  host-export allow-list — a plain env entry would have been inert.
- **`parse_to_csv.py`** — was silently dropping the first (lowest-concurrency) cell of every
  run (21 `[RUNNING]` cells -> 20 rows), tagging every MoRIIO run as `nixl`, and filing every
  MI355X run under `gfx942`. All three fixed; dropped rows now warn instead of vanishing.
- **`benchmark_niah.py`** — the accuracy check printed a score but always exited 0; now a
  real pass/fail gate, with jittered needle placement.
- Two patchers set opt-in OFF on this image (`GLM_PERSIST_GATE`, `GLM_DSA_SENTINEL_FIX`):
  both are correct on older images but here cause a C++ abort at `asm_mla.cu:945` and a
  `hipErrorIllegalAddress` respectively. The gate is by image, not by model.
- Removed a committed NFS silly-rename artifact (`.nfs0000000016f44d2b00008188`, a stale
  byte-copy of `run_xPyD_models.slurm`).

## Testing status

FP8 is the measured path — NIAH accuracy plus the 28,672/1,024 sweep at concurrency
16/32/64 above.

**`GLM-5.2-MXFP4` is config-only and has not been booted yet** — bring-up is queued. Its
recipe is the FP8 one with the two values that provably do not transfer re-derived. Both
`models.yaml` and `GLM52_MI355X.md` flag it as untested; treat its first run as a bring-up,
not a benchmark.


raviguptaamd:
## EP16 cross-node disagg + Crusoe/ionic results (additive)

Adds the EP16 (expert-parallel across 2 nodes per role) path and a three-config comparison, **without touching** the validated AAC 1P/1D-EP8 content or the NIAH harness (already correct).

**1. MoRI PR#558 for cross-node EP on ionic** (`48d62b5`)
- EP16's MoE all2all crosses the ionic fabric; GPU IBGDA doorbell fails under KVM/VFIO. Tej's PR #558 (`itej89/mori @ feat/ep-rdma-sharing`) adds a host-CPU proxy (`MORI_EP_OVER_RDMA=1`).
- Opt-in `--build-arg WITH_MORI_EP_OVER_RDMA=1` switches the MoRI source to Tej's fork and applies `docker/mori_pr558_ionic.patch` (ionic atomic-MR strip). **Default build (=0) reproduces the validated EP8 image byte-for-byte.**

**2. EP16 recipe + Crusoe/ionic results** (`99d6629`)
- `vllm_pd_ep16_launch.sh` + `ep16_orchestrate.sh` — per-node launcher + 4-node driver, handling the 4 bring-up gotchas (no `--data-parallel-start-rank` on master; `NCCL_IB_DISABLE=1` for cross-node DP; 16G MoRI heap vs util budget; GPU-drain before relaunch).
- `GLM52_EP16_IONIC.md` — recipe + why PR#558 is required.
- `RESULTS_CRUSOE_IONIC.md` + `glm52_crusoe_ionic_report.html` — three-config matrix.

**Results** (Crusoe MI355X + ionic, GLM-5.2-FP8, via vllm-router — separate cluster/shape from the AAC EP8 sweep, not merged):

| Config | TPOT | tok/s @ 8K con64 |
|--------|------|------------------|
| TP8 | ~22 ms | — (latency king) |
| EP8 | ~60 ms | 371 |
| EP16 | ~117 ms | 280 |

Each all2all hop ~doubles TPOT. **NIAH 10/10** for all three (2K/20K/50K/90K). Blocked levers documented: DBO (MoRI `supports_async=False`), MXFP4 (Triton Code-209 on gfx950), MTP (hangs in disagg).

MXFP4 fix across all cases is tracked as follow-up.

PR #206, opened 2026-08-18
vllm_dissag: GLM-5.1-FP8 (MLA+DSA) MoRI-EP WideEP disaggregated enablement

MIR-AMD:
Rebased, trimmed successor to #176 (that branch is now `CONFLICTING` against `develop`).

Adds GLM-5.1-FP8 (`GlmMoeDsaForCausalLM`, MLA + DeepSeek Sparse Attention) as a MoRI-EP wideEP disaggregated model. Validated 1P/1D EP8, 2P/2D EP16, and 4P/4D EP32 (NIAH ~93-97%; EP32 costs ~17% TPOT vs EP16 due to wider cross-node all2all).

## What lands

- New per-model image `docker/vllm_disagg_inference.glmv5.1.ubuntu.amd.Dockerfile`. The DeepSeek image is untouched. Pins: `VLLM_REF=glm5.1-dsa-wideEP_on_vllm-v0.27`, `AITER_REF=e03fa6040`, `MORI_REF=42e895472b08`, `WITH_NIXL=0`.
- Card `pyt_vllm_disagg_mori_glm-5.1-fp8` in `scripts/vllm_dissag/models.json` (per-directory catalog, post-#197).
- Recipe `GLM-5.1-FP8` in `scripts/vllm_dissag/models.yaml`: block-size 1 (the DSA sparse indexer requires it), AITER MLA on, decode `--max-num-batched-tokens 2048`, `DECODE_CUDAGRAPH_MODE=FULL_AND_PIECEWISE`.
- No runtime `.py` patchers. `connector_runtime_patch` in `connectors/moriio.sh` is a documented no-op for every model, so the MAD side is catalog and config only.
- NIAH accuracy harness selectable via `BENCHMARK_SCRIPT=niah`.

## Deliberately kept out

Relative to #176, this branch drops the lab leftovers and reverts two changes that would have affected unrelated models:

- Removed an `.nfs*` silly-rename artifact, `keepalive_bench.sh`, `niah_200k.py`, and `skills_vllm_disagg.md`.
- Reverted DeepSeek-V3's `--max-num-seqs 64 --max-model-len 32768` caps, which would have cut DSV3 context from ~163k to 32k on the untouched DSV3 image.
- Reverted the cluster-wide `connectors/moriio.env` GID/NIC pins, which applied to every moriio model. Site-specific fabric settings stay as submit-time env.

## The nine runtime patchers are removed, and here is the measurement

#176 shipped nine `apply_glm_*.py` scripts that rewrite vLLM's source inside the container at launch. On the image this recipe pins they are unnecessary, and two of them are fatal. Both claims are measured, not argued.

Running all nine against a checkout of the pinned commit modifies exactly one file, `vllm/v1/attention/backends/mla/rocm_aiter_mla_sparse.py`, by +20/-3. The other seven print "anchor not found" and write nothing, because the branch already carries that work natively: `ed374a928f` for the DSA indexer KV transfer, keyed `request_info.transfer_offsets[geometry_key]` in `moriio_engine.py` for the engine hunk, and a native `if peer_zmq is None` guard in `moriio_connector.py` for the abort guard.

Five arms on the pinned image, 1P/1D EP8, identical except for which patchers were allowed to run:

| Arm | Patchers that modified vLLM | NIAH 2k | NIAH 20k |
|---|---|---|---|
| A baseline | none | 9.3/10 | 8.0/10 (8k 9.7, 35k 10.0) |
| A2 control | none | 9.7/10 | 8.3/10 |
| B | sentinel + persistent-gate + sampling | warmup ok | engine died |
| C | sentinel + sampling | warmup ok | engine died |
| D | sentinel only | 10/10 x3 | engine died |
| E | persistent-gate only | — | engine died |

Arm A2 is the control: it ran on the same two nodes immediately after arm E crashed there, same image and same launcher tree, and completed cleanly. The hardware and the image are healthy; the patchers are what kill it. In every failing arm the worker process vanished with no Python traceback while prefilling a 22,101-token request, and the surviving ranks reported `Worker failed with error '[gloo] Read error: Connection reset by peer'` — the signature of a GPU memory access fault.

Both fatal patchers undo a deliberate decision in the pinned vLLM. `apply_glm_dsa_kernel_fix.py` flips the invalid-token sentinel back to `-1`, reverting fork commit `cda3648602`, whose message states that aiter's `mla_decode_fwd` dereferences `paged_kv_indices`, so `-1` becomes a page-aligned fault at disaggregated long context. `apply_glm_dsa_persistent_kernel_gate_fix.py` disables the persistent sparse-MLA kernel, which `4d84cb0fbc` (#47766) made unnecessary by keying metadata on per-request context.

Two further reasons not to keep them as opt-in fallbacks:

- `GLM_SKIP_PATCHERS` defaulted to `0`, and nothing in `models.yaml` or `models.json` set it, so anyone submitting without exporting it by hand got both fatal patchers on a good image. The failure only appears at long context — 2k passes perfectly, which is exactly what a smoke test would use.
- `apply_glm_aiter_sampling_oob_fix.py` reads a hardcoded path in an individual engineer's directory, `/shared_inference/ravgupta/aiter_sampling_fix_3658/sampling_patched`. That path is mounted on the cluster, so the patcher does fire, overlaying nine files into aiter's sampling kernels and purging the JIT cache. `SAMPLING_FIX_DIR` is not forwarded by the launcher, so it cannot be disabled from the submit side.

## What the vLLM ref actually is

`VLLM_REF` stays `glm5.1-dsa-wideEP_on_vllm-v0.27`, consistent with how the other `vllm_dissag` images name their vLLM ref. It is a branch, so `docker build` resolves to whatever the tip is that day, and the branch has already moved once since validation (`cda3648602` to `e8c186f71`/`d723eb305`). That mutability is a real gap and is being addressed separately rather than one Dockerfile at a time; meanwhile the header names the validated commit outright, so an auditable rebuild is one build arg away: `--build-arg VLLM_REF=d723eb305eb78d1bda0ed357b2b54cc29487221f`.

The header's description of that commit is corrected here. The branch tip `d723eb305e` is the base image's `dedbf6be8b` plus exactly 9 ROCm commits, 0 behind, verified with the GitHub compare API. `dedbf6be8b` is not the "upstream v0.27 tip" as previously stated — it is a `main` commit 270 ahead of the `v0.27.0` tag and 17 behind `releases/v0.27.0`, so the "v0.27" in the branch name labels this line of work rather than a release checkout.

## Which image to serve this on

The recipe comment previously named a pre-v0.27 tag as "validated". That is corrected here to `rocmshared/pytorch-private:glm5.1-vllm027-b8`, the image the Dockerfile builds. This is not cosmetic: serving the recipe on the older `rocm/pytorch-private:glm-dockerimage-built-09072026` boots cleanly and then returns empty completions (NIAH 2k scored **0/10**, no crash), which is an easy trap to fall into.

## Test plan

- [x] `scripts/vllm_dissag/tests/gate_check.sh`: 24 passed, 0 failed, including `REQUIRED_FILES` completeness after the patcher removal.
- [x] `scripts/vllm_dissag/tests/argv_assert.sh`: 30 passed, 0 failed.
- [x] `DRY_RUN=1` for GLM-5.1-FP8 emits expert-parallel, `--block-size 1`, `FULL_AND_PIECEWISE`, decode `--max-num-batched-tokens 2048`.
- [x] `DRY_RUN=1` for DeepSeek-V3 is unchanged versus `develop`.
- [x] OCI 1P/1D (EP8) serving smoke on `amd-rccl`, 2 nodes, `rocmshared/pytorch-private:glm5.1-vllm027-b8`: prefill + decode engines ready, router registration complete, disaggregated KV transfer verified end to end, **NIAH 2k retrieval 10/10**, no `EngineDeadError`, no `hipErrorIllegalAddress`, no `asm_mla.cu:945`.
- [x] Long-context NIAH sweep, 3 seeds per length: 2k **9.3/10**, 8k **9.7/10**, 20k **8.0/10**, 35k **10.0/10**, zero fault signatures.
- [x] Five-arm patcher A/B above, which is what justifies removing the patchers rather than defaulting them off.
- [x] `use_inductor_graph_partition` A/B (slurm `217352.91` off vs `.108` on): NIAH 2k **10/10 both arms**, no `unknown parameter type` in any log, matched 1024/1024 con=8 sweep mean TPOT 42.57 ms off vs 42.11 ms on. The flag is kept because it is what the published numbers were measured with, not because GLM requires it.
- [x] Per-shape warmup opt-in, matched 1024/1024 con=8 arms on one allocation (`217352`): with nothing exported the recipe's `SHAPE_WARMUP=1` reached the container and the warmup ran (mean TPOT **42.13 ms**); with `SHAPE_WARMUP=0` exported it was suppressed (**42.05 ms**); against the `igp_on_bench` baseline of **42.11 ms**, 16/16 requests and `exit 0` in every arm. The script default is off, so no other model's sweep path changes.
- [x] **Six-length single-seed NIAH reproduction of #176's accuracy claim, on this branch, with no fabric variables exported** — the tree exactly as checked out, so these numbers are reproducible from the branch itself (jobs `217918` EP8 / `217919` EP16, both on `amd-rccl`; image `sha256:3327782d4286`; weights read from the node-local NVMe copy; MoRI at the shipped `MORI_IB_GID_INDEX=1`; both `exit 0` in ~1h01m; zero hits for `EngineDeadError`, `hipErrorIllegalAddress`, `asm_mla.cu:945`, `Memory access fault`, `Traceback`):

| words | 1P/1D EP8 | 2P/2D EP16 |
|---|---|---|
| 2 000 | 10/10 | 10/10 |
| 8 000 | 10/10 | **9/10** (`octopus`) |
| 16 000 | 10/10 | **8/10** (`penguin`, `tiger`) |
| 20 000 | **9/10** (`tiger`) | **8/10** (`rhinoceros`, `tiger`) |
| 28 000 | **9/10** (`tiger`) | **8/10** (`penguin`, `tiger`) |
| 35 000 | 10/10 | 10/10 |

  Protocol is `NIAH_SEEDS=0` with warmup on, matching #176's single-seed protocol so the two are comparable. #176 claims 10/10 at all six lengths on EP8 and across 2k–35k on EP16; **that does not reproduce.** EP8 diverges by one needle at 20k and 28k; EP16 diverges by one to two needles at every length from 8k through 28k. Both topologies recover to 10/10 at 35k, so this is not monotonic long-context collapse, and neither run showed a fault or a crash. `tiger` accounts for 4 of the 5 sub-10 cells, across two topologies and independently of seed, which points at a needle-position effect rather than seed noise. It is also not a regression introduced here: no arm has ever measured 16k or 28k with the patchers present, so there is no earlier number on this harness to have regressed from.

2P/2D EP16 is measured on this branch (job `217919`, table above) rather than inherited from #176. It runs clean — both prefill and both decode engines up, router registration complete, disaggregated KV transfer confirmed by distinct prefill/decode addresses on the served completion, `exit 0`, no fault signatures — but its retrieval accuracy sits below EP8 at every length from 8k to 28k, so the earlier "validated, 10/10 across 2k–35k" claim is not restated.

**4P/4D EP32 is now validated** (raviguptaamd). The root cause of the earlier corruption was `mori.py combine()` receiving post-dispatch `topk_ids` instead of pre-dispatch — harmless at EP8/EP16, corrupt at EP32. Fix is in the pinned vLLM (`623fdc946b`). Measured on the current branch (job `217919`, image `glm5.1-pr206-9f16198`): NIAH 2k **9.7/10**, 8k **9.3/10** (3 seeds). Perf: 8192/1024 con=32 TPOT **70.4 ms** vs EP16's **60.0 ms** (~17% cost). Zero GPU faults, 100% request success. The cross-node all2all inside the captured decode step does not compress as EP widens — EP32 buys capacity, not per-token latency.

### Why the NIAH harness is in this PR

`benchmark_niah.{py,sh}` already exist in `develop` but are unreachable: the `BENCHMARK_SCRIPT` selector accepts only `sweep` and `long_context`, so no model can invoke them. This PR adds the `niah)` case plus the harness work needed to make GLM's accuracy claims reproducible, since the long-context retrieval numbers quoted in `models.yaml` are produced by this script:

- per-length warmup inside the harness (`NIAH_WARMUP`, a separate knob from the sweep's `SHAPE_WARMUP`), so a cold compile is recorded as a compile stall rather than a wrong answer;
- seeded multi-run scoring, reported as mean/min/max rather than one run;
- a router readiness probe in place of a blind `sleep 10`;
- a timeout-versus-wrong-answer distinction, so a compile stall is not recorded as 0/10.

To be precise about one of these: `enable_thinking: False` is defensive, not a fix for a diagnosed bug. `develop`'s harness already scored `content + reasoning_content`, so the flag only prevents chain-of-thought from consuming `max_tokens` before the answer is emitted.

The one change on a path other models share is the per-shape warmup in `benchmark_xPyD.sh`, and it defaults to **off** (`SHAPE_WARMUP=0`). GLM-5.1-FP8 opts in through its `models.yaml` `env:` block, which is the configuration its published TPOT was measured under. No existing model's benchmark path changes.


MIR-AMD:
## OCI 1P/1D EP8 NIAH-2k smoke — **FAIL** (2k retrieval 0/10)

Slurm job `216847` on partition `amd-rccl`, 2 nodes (`useocpm2m-097-121` prefill+proxy, `useocpm2m-097-144` decode). GLM-5.1-FP8 disaggregated 1P/1D EP8 over the MoRI-EP wideEP path (`CONNECTOR=moriio WIDE_EP=1 EP_BACKEND=mori`, `GLM_SKIP_PATCHERS=1`), NIAH harness at a single 2000-word context, one seed, warmup enabled (`BENCHMARK_SCRIPT=niah NIAH_WORDS=2000 NIAH_SEEDS=0 NIAH_WARMUP=1`).

- **Image:** `rocm/pytorch-private:glm-dockerimage-built-09072026`
- **Model:** `✓ Selected MODEL_PATH: /mnt/m2m_nobackup/models_blog/GLM-5.1-FP8 (available on all nodes)`

### Decisive lines

```
[glm] GLM_SKIP_PATCHERS=1: image carries DSA fixes in-source; skipping runtime patchers.
Proxy (vllm_router) ready for benchmarking on useocpm2m-097-121:10.158.212.154:30000
Router registration complete after 5s.
=== NIAH warmup (one throwaway request per length) ===
words=  2000  [warmup] ok
words=  2000  seed=0  found= 0/10  []
=== NIAH summary (mean/min/max across 1 seed(s)) ===
  words=  2000  mean=0.0/10  min=0  max=0  (n=1)
```

### Verdict: FAIL

Bring-up is clean and the patcher-skip path works exactly as intended, but the scored retrieval is **0/10 at 2000 words**, which is the one bar item that fails.

**What passed:**

- `GLM_SKIP_PATCHERS=1` logged on both nodes, and **no `apply_glm_*.py` was applied** — the patcher filenames that appear in the logs are only the `Listing NIXL_COOKBOOK_PATH: /opt/nixl-vllm-cookbook` directory listing, not an application step.
- Proxy ready and router registration completed in 5s. The bring-up `/v1/completions` canary and both NIAH requests (warmup + scored) all returned HTTP 200 through the P/D path, with `MoRIIO handshake done for request chatcmpl-...` on the prefill side.
- No `EngineDeadError`, `hipErrorIllegalAddress`, `asm_mla.cu:945`, `Traceback`, or `CUDA error` anywhere in `slurm-216847.out` (`.err` is 0 bytes) or in any per-node log.
- Clean exit, not a wall-clock kill: `Script completed successfully.` and `sacct` reports `COMPLETED  00:56:38  0:0`, well inside the 90-minute limit.

**Failing evidence:** `words=  2000  seed=0  found= 0/10  []`. The scored request returned HTTP 200 in roughly two seconds and its `content` / `reasoning_content` contained none of the 10 planted animals, so this is a real wrong answer rather than a timeout or a cold-compile artifact — warmup for this shape had already succeeded (`[warmup] ok`) immediately before. At 2000 words this is far below the ~14–18k-token threshold where the upstream sparse-MLA prefill indexer corruption is documented, so the known long-context caveat does not explain it. Consistent with degraded output, the bring-up canary was also degenerate: prompt `Who is AMD CEO?` returned `" Who is the CEO of AMD? Who is the"`.

**Noted but not bar-failing:** `[niah] WARN: router readiness not confirmed in 300s; proceeding (warmup will absorb)` — the harness `/v1/models` readiness probe never observed a 200 from `vllm-router` even though the router was already serving traffic on the same port, so the run burned the full 300s gate before scoring. Worth a follow-up on the probe endpoint.

This run exercises the trimmed patcher defaults from this PR (persist-gate and DSA sentinel both opt-in OFF) on the v0.27 image (server reports `vllm-0.16.0rc2.dev4561+g545292d14-dp8-ep-07541063`).

Logs: `/shared_inference/miali/model_blog_logs/slurm-216847.out` and `/shared_inference/miali/model_blog_logs/216847/`

MIR-AMD:
Serving smoke is green, so the test plan above is now complete.

OCI `amd-rccl`, 2 nodes, 1P/1D (EP8), image `rocmshared/pytorch-private:glm5.1-vllm027-b8`:

- `[glm] GLM_SKIP_PATCHERS=1: image carries DSA fixes in-source; skipping runtime patchers.`
- prefill + decode engines ready, `Proxy (vllm_router) ready`, router registration complete after 5s
- disaggregated KV transfer verified end to end
- **NIAH 2k retrieval 10/10**
- no `EngineDeadError`, no `hipErrorIllegalAddress`, no `asm_mla.cu:945`; job exited cleanly

One note worth flagging for anyone reproducing this: the same branch on the older `rocm/pytorch-private:glm-dockerimage-built-09072026` boots fine and then returns empty completions (NIAH 2k = 0/10) with no crash. That is why the recipe comment now names the v0.27 image explicitly.

raviguptaamd:
Hi @MIR-AMD — sharing some 4P/4D EP32 results in case they're useful for this PR, since the header currently lists EP32 as an open defect.

**EP32 appears to be fixed.** The garbage-at-all-lengths behavior traces to `MoriPrepareAndFinalize.finalize()` passing the *post-dispatch* `topk_ids` to `mori_op.combine()`, so combine gathers expert partials from the wrong nodes (harmless at EP8/EP16, corrupt at EP32). Fix (Wei Sun's, `e25bf182`) stashes the original pre-dispatch ids in `prepare()` and uses them in `combine()`. It's now folded into the branch this image builds:
- `raviguptaamd/vllm@623fdc946b` on `glm5.1-dsa-wideEP_on_vllm-v0.27` (`mori.py`, +16/-1).

**Measured, 4P/4D EP32, NIAH 8k (single-needle, seed 0):**

| | Baseline (no fix) | With fix |
|---|---|---|
| NIAH 8k | 0/10 (empty) | **10/10** |
| bring-up canary | garbage / repetition | coherent, correct answer |

Perf-neutral at 8192/1024 con=32 (TPOT 74.0 vs 74.1 ms; output 380.99 vs 380.98 tok/s). No crashes either arm — it's a pure correctness bug.

Since this branch already carries the commit, a rebuild picks it up. A couple of doc spots that may want a follow-up when convenient:
- Header (line ~44): "4P/4D EP32 is a KNOWN OPEN DEFECT..." — no longer matches these runs.
- Auditable pin (line ~181): `--build-arg VLLM_REF=d723eb305e...` points at the pre-fix commit; `623fdc946b` is the one these numbers were measured on.
- PR description: still says "EP32 remains a known defect and is not claimed" — could be updated to reflect the fix when you get a chance.

Happy to share full logs/job IDs if helpful. Not urgent — just flagging so the image contract stays accurate.

raviguptaamd:
### Minimal Dockerfile change to move to a fresher, faster GLM-5.1 stack (optional follow-up)

Sharing a validated path to bump the pins ~2.5 months forward (aiter/MoRI) while keeping the recipe's cudagraph decode. **One file changes — `docker/vllm_disagg_inference.glmv5.1.ubuntu.amd.Dockerfile`, 4 build-args.** No `models.yaml` change (the existing `DECODE_CUDAGRAPH_MODE: FULL_AND_PIECEWISE` + persistent MLA + `mori_low_latency` config is exactly what validated). Router pin unchanged.

```diff
- ARG MORI_REF=42e895472b08
+ ARG MORI_REF=624002c897a3

- ARG AITER_REPO=https://github.com/ROCm/aiter.git
- ARG AITER_REF=e03fa6040
+ ARG AITER_REPO=https://github.com/raviguptaamd/aiter.git   # until ROCm/aiter#4957 merges, then revert repo
+ ARG AITER_REF=624e43586b                                    # SHA-pinned (ROCm/aiter#4957: gqa64 gfx942 fix)

-         pip install --no-deps -U "flydsl>=0.1.7,<0.1.9" && \
+         pip install --no-deps -U "flydsl==0.3.1" && \

- ARG VLLM_REF=glm5.1-dsa-wideEP_on_vllm-v0.27
+ ARG VLLM_REF=094820b5d   # SHA-pinned; raviguptaamd/vllm branch glm5.1-dsa-wideEP_on_d626108b
```

**Why the aiter bump needs the fork ref for now:** newer stock aiter (>= 1d872fa) reintroduces a gfx942 gqa64 sparse-MLA decode fault (`mla_a8w8_qh64_qseqlen1_gqaratio64_v3_ps` GPU-faults). The fix is a 7-line change filed upstream as **ROCm/aiter#4957** — once it merges, `AITER_REPO` reverts to `ROCm/aiter.git` at the merged SHA.

**What the vLLM bump is:** the same GLM DSA commits (incl. the EP32 combine-topk fix `623fdc946b`) cherry-picked cleanly onto upstream vLLM `d626108b` (2026-08-20).

**Validated** (MI300X/gfx942, MoRI-EP WideEP disagg, cudagraph decode on):
- NIAH: 1P/1D EP8 — 2k 10/10, 8k 9/10; **2P/2D EP16 — 2k 10/10, 8k 10/10**.
- Perf 2P/2D EP16 (TPOT median / total tok·s⁻¹, 100% success):
  - 8192/1024  con32: 59.5 ms / 4001 ; con64: 64.0 ms / 6527
  - 16384/1024 con32: 59.7 ms / 6723 ; con64: 66.0 ms / 10931

TPOT ~59–66 ms vs the current stack's ~74 ms at the same shape — the fresh stack is faster with cudagraph decode intact. (4P/4D EP32 not yet re-run on the fresh stack — needs an 8-node alloc.)

Not urgent — flagging the exact minimal change so it's a small, auditable bump when convenient.

raviguptaamd:
@MIR-AMD — I pushed one commit directly to `glm5.1-wideep-mad` (`094d108`), touching only `docker/vllm_disagg_inference.glmv5.1.ubuntu.amd.Dockerfile`. Happy to back it out or reshape it if you'd rather take it differently. What it does and why:

### Pins

| | before | after |
|---|---|---|
| MoRI | `42e895472b08` | `624002c897a3` |
| AITER | `ROCm/aiter@e03fa6040` | `raviguptaamd/aiter@624e43586b` |
| flydsl | `>=0.1.7,<0.1.9` | `==0.3.1` |
| vLLM | `glm5.1-dsa-wideEP_on_vllm-v0.27` (branch) | `094820b5d` (sha) |
| router | `ravgupta/discovery-dp-rank-roundrobin` | `ravgupta/dp-roundrobin-on-tip` |

### The EP32 defect is fixed

The header called 4P/4D EP32 a KNOWN OPEN DEFECT (token corruption at all context lengths). Root cause was in `fused_moe/prepare_finalize/mori.py`: `combine()` was being handed the **dispatched** `topk_ids` rather than the **original** ones, so expert outputs were scattered back against the wrong routing at EP32. Fix stashes the original in `prepare()` and uses it in `combine()` (vLLM `623fdc946b`, now in the pinned vLLM). Measured at 4P/4D EP32: **NIAH 8k 0/10 → 10/10, perf-neutral.**

### Why AITER needed a fork (temporary)

This one is the reason the old pins couldn't simply be bumped. Newer aiter claims native gfx942 support for gqa64 fp8 decode and routes GLM-5.1 to `mla_a8w8_qh64_qseqlen1_gqaratio64_v3_ps`, which **GPU-faults**. That's the fault the old comment attributed to "aiter > e03fa6040" generally — it's actually this one clause. The fix is 7 lines: let gqa64 fall through to aiter's capture-safe persistent view-fold, which keeps **cudagraph decode** working (the obvious alternative — dropping to eager decode — costs far too much TPOT to accept).

Filed upstream as **ROCm/aiter#4957**. I know a fork ref inside a ROCm MAD recipe isn't ideal — the commit and the header both mark it TEMPORARY with an explicit instruction to revert `AITER_REPO` to `ROCm/aiter` once #4957 merges. If you'd prefer to hold the whole bump until then, that's reasonable too; the alternative is staying on `e03fa6040` and not getting the EP32 fix. MoRI `624002` and flydsl `0.3.1` are co-requisites of the aiter bump (`flydsl_fp8_mqa_logits` doesn't exist on 0.1.x).

### Validation — and what is NOT covered

Config: prefill MoRI-HT eager + decode MoRI-LL `FULL_AND_PIECEWISE`.

- 1P/1D EP8 and 2P/2D EP16 on the bumped pins: NIAH 2k **10/10**, 8k **10/10**; 8192/1024 con32 TPOT **~59 ms**
- The EP32 0/10 → 10/10 result above was measured on the **pre-bump** stack

**4P/4D EP32 has not been re-validated on the bumped pins** — I've only had a 5-node allocation and it needs 8. The header now says this explicitly rather than implying coverage. I'll post numbers when I get the nodes.

### Two smaller things

- `VLLM_REF` was a **branch name**, so two builds on different days could ship different engines. Now pinned to a sha; `/app/versions.txt` still records what resolved.
- Refreshed the header comments that described the old pins (MoRI/AITER/vLLM rationale text) so they match the new ARGs — that's most of the line count in the diff.

One request on the PR description: it still describes EP32 as broken and recommends 1P/1D and 2P/2D only. Worth updating now that it's fixed.

raviguptaamd:
@MIR-AMD — pushed two commits and synced the branch with `develop` (now **behind 0**). The `develop` merge brings only CODEOWNERS + the Atom Dockerfile, disjoint from this PR's files, no conflicts.

## Validation: 4P/4D EP32 now measured

I built an image from this PR's head **unmodified** (`glm5.1-pr206-9f16198`) and ran the full matrix. Previous numbers came from images that predated the recipe; this is the first from the committed artifact.

**NIAH, 3 seeds:**

| topology | 2k | 8k |
|---|---|---|
| 2P/2D EP16 | 9.7/10 | 9.7/10 |
| 4P/4D EP32 | 9.7/10 | 9.3/10 |

**Perf (TPOT median / output tok/s):**

| shape | con | 2P/2D | 4P/4D |
|---|---|---|---|
| 8192/1024 | 32 | 60.0 ms / 440 | 70.4 ms / 386 |
| 8192/1024 | 64 | 64.7 ms / 765 | 74.4 ms / 640 |
| 16384/1024 | 32 | 60.9 ms / 388 | 70.6 ms / 346 |
| 16384/1024 | 64 | 66.8 ms / 640 | 75.6 ms / 583 |

**0 failed requests across all 8 cells, 0 GPU faults.** So the STATUS block's "not re-validated on the bumped pins" is replaced with these numbers.

Two things I'd rather state than let a reader infer:

- **The 9/10s are seed variance, not length collapse.** Across 12 cells every score is 9 or 10, no pattern by topology or context length. Reads as ~93–97% retrieval — I did not want the header claiming a clean 10/10 off single-seed runs.
- **EP32 costs ~17% TPOT over EP16.** The cross-node all2all inside the captured decode step doesn't compress as EP widens. 4P/4D buys capacity, not per-token latency, and the header now says so.

## The one functional fix — and it's outside the Dockerfile

**4P/4D could not boot on this recipe.** `GPU_MEMORY_UTILIZATION=0.8` OOMs during EngineCore init at EP32:

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 6.00 GiB.
GPU 0 has 191.98 GiB total, 6.95 GiB free, 157.95 GiB allocated by PyTorch.
```

That SIGTERMs every rank, and then the socket barrier prints `Waiting for nodes` for ~30 minutes — it reads as a hang, not a crash. Cost me a while to spot. The MoRI dispatch/combine buffers scale with EP width but the memory fraction didn't.

The default is now derived from decode EP width in `run_xPyD_models.slurm`: **0.70 at EP>=32, 0.8 below**. 0.70 boots (66.79 GiB KV cache) and holds through con=64; 2P/2D at 0.8 is untouched — I validated that at stock in the same session, so this is conditional, not a blanket lowering. Submit-time `-e` still overrides both.

**Heads-up on two things you may want to push back on:**

1. **This is the only change outside the glmv5.1 Dockerfile.** I know that widens the PR's scope. The alternative was leaving a recipe that claims EP32 works and can't launch it, which seemed worse — but it's your PR, so say the word and I'll pull it.
2. **`models.yaml`'s `GPU_MEMORY_UTILIZATION` key is inert.** The `-e` on the docker run line is unconditional, and `vllm_disagg.sh` skips any yaml key already in the environment — so that key has never taken effect. I left it (it documents intent) and added a comment pointing at the live value. Worth knowing if anyone has been tuning it there expecting results.

## Also

`ROUTER_REF` was a **branch name**, so two builds on different days could ship different routers — the same mutable-ref hazard the `VLLM_REF` sha pin removed. Pinned to `82dc9811`. Its comment still described the old PR #181 branch and its two commits; upstream has since merged the round-robin fix (`prefill_dp_round_robin`), so only the KV-notify fix is carried locally now.

## Still open

**The aiter pin is still a fork** (`raviguptaamd/aiter@624e43586b`). ROCm/aiter#4957 is open with no review comments yet. Until it merges this PR carries a fork dependency; the Dockerfile marks it TEMPORARY with a revert instruction. Nothing we can do from this side.

Worth being straight about pin freshness too: these are the **newest pins that actually run GLM DSA**, not tip-of-tree. vLLM is ~180 commits behind upstream main, aiter ~49, MoRI 5. I tried tip four times and every attempt died — first a flydsl import error, then GPU memfaults on the DSA decode kernel. The kernel ABI is co-tuned between a specific vLLM commit and a specific aiter; `d626108b` + `1d872fa` is the pair that works.

Cemberk:
Three design issues on the MAD integration side. The model enablement itself looks fine.

**1. NIAH has no card and no result rows.**

`run_xPyD_models.slurm:438` adds the `niah)` selector case, but nothing reaches it. The new entry at `models.json:282` sets `BENCHMARK_COMBINATIONS: "1024/1024"` and no `BENCHMARK_SCRIPT`, so it runs the default sweep — `BENCHMARK_SCRIPT=niah` only happens if someone exports it by hand. And `perf.csv` is written only by `benchmark_xPyD.sh:107-108` via `parse_to_csv.py --perf-csv`; neither `benchmark_niah.sh` nor `benchmark_niah.py` calls it, so a NIAH run produces zero madengine rows.

Not schedulable by tag, can't regress in CI, nothing on the dashboard.

Fix: add a second card in `scripts/vllm_dissag/models.json` — `pyt_vllm_disagg_mori_glm-5.1-fp8_niah`, same image and topology, `"BENCHMARK_SCRIPT": "niah"` in `env_vars`, plus a `niah` tag. Then make `benchmark_niah.sh` emit perf.csv rows through `parse_to_csv.py` (retrieval score, one row per length) — the tag alone still reports nothing.

**2. `GPU_MEMORY_UTILIZATION` in `models.yaml` is inert.**

`vllm_disagg.sh:171` skips any yaml key already in the environment; `run_xPyD_models.slurm:644` forwards `-e GPU_MEMORY_UTILIZATION=...` unconditionally. So the yaml value never applies. It's the only one of the 36 `env:` keys affected, and it predates this branch — `Qwen3-30B-A3B` sets it at `models.yaml:146` with no effect.

This PR adds a second dead instance at `models.yaml:222` and documents the trap in a slurm comment rather than resolving it, while making the number load-bearing via the EP32 tuning. One value, three declarations: dead yaml `0.80`, slurm `0.8`, slurm EP32 `0.70`.

Fix, either direction:
- slurm authoritative — drop the key from `models.yaml:146`, `:222`, and the header list at `:12-18` that advertises it.
- yaml authoritative — `${GPU_MEMORY_UTILIZATION:+-e GPU_MEMORY_UTILIZATION=$GPU_MEMORY_UTILIZATION}` at `:644`, with the EP-width default resolved after the yaml load.

Note the second one breaks EP32 boot if done naively: GLM's `0.80` would override the `0.70` that exists to stop the EngineCore OOM.

**3. The EP-width memory default is global, not scoped to GLM.**

`run_xPyD_models.slurm:553-554` keys off topology alone and applies to every model in the directory. No declared card reaches EP32 today (all `-N 2 -n 2`), but a hand-run DeepSeek-V3 at 4P/4D silently goes `0.8` → `0.70`, changing its KV cache size and its numbers. The `0.70` was measured on the GLM stack.

Fix: scope it to GLM like the other GLM knobs, or keep it global deliberately and say so in the README model table.

Cemberk (approved):
LGTM 👍

PR #211, opened 2026-08-18
Sync mad-rccl with develop (resolves models.json migration conflict from #210)

i-kosarev:
Supersedes the conflicting merge in #210.

## Why #210 conflicts

Both branches carry the same cherry-picked commits (#195, #196, #198, #201, #203) under different SHAs, so the only real divergence is:

- **develop**: #197 *"Migrated models from global to per directory"* — the root `models.json` was split into `scripts/<dir>/models.json`.
- **mad-rccl**: #194 (multi-node scaleout benchmarks + mad-slurm-multinode skill) and #200 (sglang-disagg QP fix / optional overlay stages), which kept adding entries to the **root** `models.json`.

Hence a single conflicted file: `models.json`.

## How it is resolved

Resolution follows develop's migration rather than reverting it:

- root `models.json` takes develop's post-migration content — only `primus_train`, which must stay at the root because `scripts/primus_train/` has a `get_models_json.py` (per the discovery rule documented in `tools/migrate_to_dir_models.py`);
- the four entries that existed **only** on mad-rccl were migrated into per-directory files using that tool's own translation rules (`dockerfile` → `../../docker/...`, `scripts` → path relative to the script dir), so they match the convention of the already-migrated entries:
  - `scripts/primus_megatron-lm/models.json` (new file) — `primus_pyt_megatron_lm_train_llama-3.1-{8b,70b,405b}_overlay`;
  - `scripts/sglang_disagg/models.json` — `sglang-disagg-deepseek-r1-overlay` appended to develop's 16 entries.

## Verification

- Union of model names across all `models.json` files is **exactly** the union of both branches (**137**) — nothing dropped from either side (checked programmatically both ways).
- Every `models.json` parses; no conflict markers remain.
- `scripts` paths of the migrated entries resolve to existing files; both overlay Dockerfiles exist (`docker/primus_megatron_train_rccl_overlay.ubuntu.amd.Dockerfile`, `docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile`).

## Test plan

- [ ] `madengine` discovers the three `primus_pyt_megatron_lm_train_llama-3.1-*_overlay` models from `scripts/primus_megatron-lm/models.json`
- [ ] `madengine` discovers `sglang-disagg-deepseek-r1-overlay` from `scripts/sglang_disagg/models.json`
- [ ] existing develop-migrated models still resolve unchanged

Made with [Cursor](https://cursor.com)


mkuznet1 (approved):
lgtm

PR #213, opened 2026-08-19
Miali/agentic suite

MIR-AMD:
## Summary

Adds **AgentX** — a config-driven, multi-workload agentic trace-replay benchmarking suite for disaggregated (prefill/decode) serving — and wires it into both the **SGLang** and **vLLM** disaggregated launchers as a recipe-native option (`AGENTIC=1`).

The suite replays real multi-turn agentic traces (instead of a random ISL/OSL sweep) to exercise prefix-cache reuse, and reports cache-hit / throughput metrics. The core is backend-agnostic (speaks the OpenAI API only); each backend integration is a thin (~50-60 line) launcher hook.

## What's included

**Core suite (`scripts/common/`, backend-agnostic)**
- `agentx/agentx_config.py` — YAML + env config resolution, preset inheritance (with circular-preset detection), config-less `AGENTIC_WORKLOAD` shorthand.
- `agentx/gen_agentx_profile.py` — deterministic corpus generation from a workload profile (seeded, reproducible).
- `agentx/verify_agentx_profile.py` — 13-axis verification pre-gate (ISL/OSL/Turns/Delay P50/P90/P99 + Cache-hit P50) against tolerance bands.
- `agentx/filter_weka_corpus.py` — Tier-2 corpus filtering (`--max-turns`, `--max-isl`, `--sample`).
- `benchmark_agentic_suite.sh` — multi-workload driver (per-workload corpus materialization, concurrency sweeps, `suite_summary.json`, continue-on-failure).
- `agentic_lib.sh` — shared library (venv setup, corpus materialization, context gating, router readiness, replay command build, result/rollback handling).
- `preflight_weights.sh`, `validate_agentic_result.sh`, `verify_agentic_image.sh` — supporting checks.
- Profiles: `small`, `conformance_256k`, `conformance_512k`, `inferencex_256k`, plus `custom.example.yaml`.
- Docs: `agentx/README.md` (schema + flow), `agentx/HOW_IT_WORKS.md` (mechanism walkthrough), `agentx/profiles/README.md` (authoring guide), `agentx/SCENARIOS.md` (scenario cookbook).

**SGLang disagg integration (`scripts/sglang_disagg/`)**
- `benchmark_agentic.sh` hook (selected via `AGENTIC=1`; router port `:2322`).
- Suite env-var forwarding in `run_xPyD_models.slurm`; radix prefix-cache enablement on the agentic path; README invocation examples.

**vLLM disagg integration (`scripts/vllm_dissag/`)**
- `benchmark_agentic.sh` hook — `AGENTIC_PORT` resolves to the vLLM proxy `BENCHMARK_PORT` (vs SGLang's fixed router port).
- `agentic_models_shim.py` — synthesizes `/v1/models` (readiness + model resolution) because the PD router 503s on `/v1/models` under MoRIIO discovery.
- Connector edits (`moriio.sh`, `rixl.sh`) — strip `--no-enable-prefix-caching` and export backend `/metrics` endpoints (`AGENTIC_SERVER_METRICS`) on the agentic path.
- Suite env-var forwarding in `run_xPyD_models.slurm`; README invocation examples.

All backend edits are gated/default-off so non-agentic sweeps are byte-for-byte unchanged.

## Enhancements since the initial revision

- **Context auto-discovery** — the served `max_model_len` is now resolved from the real prefill worker (`:3000/v1/models`, with a `/get_server_info` fallback on SGLang) and fails fast if unresolvable, so `MAX_MODEL_LEN` no longer needs pinning.
- **`source: corpus`** replay of an existing on-disk weka_trace corpus, and a **configurable aiperf `--scenario`** (`run.scenario` / `AGENTIC_SCENARIO`, per-workload override).
- **Config precedence fix** — `run.concurrency`/`run.duration` from the YAML are now honored (precedence: explicit env > config file > loader defaults); the context clamp-and-run notice is logged as `[agentic][WARN]` instead of `[agentic][ERROR]`.
- Review hardening — file-handle scoping, required-field/CLI-arg guards, launcher IP-resolution fail-fast, MoRI preflight, and a safe-by-default `AGENTIC_ALLOW_UV_INSTALL` gate.

## Usage

```bash
# multi-workload suite
AGENTIC=1 AGENTIC_CONFIG=scripts/common/agentx/agentic.example.yaml sbatch ... run_xPyD_models.slurm
# single-workload shorthand
AGENTIC=1 AGENTIC_WORKLOAD=small sbatch ... run_xPyD_models.slurm
```
See `scripts/common/agentx/README.md` and each launcher README for the full env-var table and examples.

## Testing

Validated on the OCI `amd-rccl` cluster with **Qwen3-32B** (dense) and **DeepSeek-V3 / DeepSeek-V3-5layer** (MoE). The served context window is now auto-discovered (40960 for Qwen3-32B, 163840 for DeepSeek-V3) — no `MAX_MODEL_LEN` pin required.

- **SGLang: fully validated end-to-end on GPU** — MoRI+TP (dense) and MoRI DP1 wideEP (MoE). Real completions, `error_rate 0.0`. Multi-workload suite iterates all workloads and writes the top-level `suite_summary.json` aggregate; `gpu_cache_hit_rate` is populated (e.g. 36.4% / 23.5% / 40.8% across a 3-workload run). Live HF-download replay (weka 256k corpus + Tier-2 filter, `max_isl=150000`) runs with successful requests (98/100 traces kept, `error_rate 0.0`).
- **vLLM: validated end-to-end on GPU** — streaming chat over the MoRIIO PD path now works on the rebuilt image; rixl+TP (dense) and moriio wideEP (MoE) run with `error_rate ~0`. The `/v1/models` shim, metrics export, corpus gen, 13/13 pre-gate, `RUN_INVALID` rollback, and continue-on-failure are all exercised.
- **Node-free suite** (`scripts/common/agentx/tests/run_offline.sh`) passes 30/30, and both `resolve_ctx_offline.sh` context-discovery suites (vLLM + SGLang) pass all cases.

## Known limitations / follow-ups

Resolved since the initial revision:
- ~~`max_model_len` autodetect returns 0 against the router~~ — **RESOLVED**: auto-discovered from the real prefill worker (`:3000/v1/models`, `/get_server_info` fallback); verified node-free and live (40960 / 163840). No pin needed.
- ~~SGLang co-located prefill launched without `--enable-metrics`~~ — **RESOLVED / clarified**: the agentic path enables `--enable-metrics --enable-metrics-for-all-schedulers` on **both** prefill and decode and points aiperf at both worker `:3000` endpoints, so `gpu_cache_hit_rate` is populated when a run completes a measured window.
- ~~vLLM GPU replay needs a synced MoRIIO image~~ — **RESOLVED**: rebuilt image with the MoRI stack; MoRIIO PD streaming chat works end-to-end.
- ~~Live HF trace replay validated by dry-run only~~ — **RESOLVED (SGLang)**: live HF-download replay ran on DeepSeek-V3 with a Tier-2 filter, `error_rate 0.0`.


scripts/vllm_dissag/benchmark_agentic.sh
```diff
@@ -0,0 +1,101 @@
...
+        _mml=""
+        for _i in 1 2 3; do
+            _mml="$(curl -sf "http://${_worker}/v1/models" 2>/dev/null \
+                    | python3 -c 'import sys,json
+try:
+    d=json.load(sys.stdin)
+except Exception:
+    print(""); sys.exit()
+data=d.get("data") or []
+print((data[0].get("max_model_len") if data else "") or "")' 2>/dev/null)"
```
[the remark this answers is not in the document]
MIR-AMD: Resolved — false positive. The `python3 -c '...'` continuation lines here (`try:`, `data=`, `print(...)`) start at **column 0**, not indented with the surrounding bash; only the block bodies are indented 4 spaces, so the emitted Python is valid. Empirically confirmed by `scripts/vllm_dissag/tests/resolve_ctx_offline.sh`, which drives this exact resolver against a stub `/v1/models` and passes — an IndentationError would blank the parser output and trip the `could not resolve served max_model_len ... pin MAX_MODEL_LEN` die-path.

scripts/sglang_disagg/benchmark_agentic.sh
```diff
@@ -0,0 +1,113 @@
...
+    data = d.get("data") or []
+    if data:
+        v = data[0].get("max_model_len") or ""
+else:
+    sa = d.get("server_args") or {}
+    v = d.get("max_model_len") or d.get("context_length") \
+        or sa.get("max_model_len") or sa.get("context_length") or ""
+print(v or "")
+PY
+)" "$_kind")"
```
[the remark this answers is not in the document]
MIR-AMD: Resolved — false positive. This uses a `<<'PY'` heredoc whose lines are written at **column 0** (heredoc content is literal and not indented into the bash block), so the generated Python is valid. Confirmed by `scripts/sglang_disagg/tests/resolve_ctx_offline.sh`, which resolves `131072` from a stub worker through this path (a real IndentationError would fail that test).

scripts/common/agentx/templates/benchmark_agentic.template.sh
```diff
@@ -0,0 +1,127 @@
...
+        _mml=""
+        for _i in 1 2 3; do
+            _mml="$(curl -sf "http://${_worker}/v1/models" 2>/dev/null \
+                    | python3 -c 'import sys,json
+try:
+    d=json.load(sys.stdin)
+except Exception:
+    print(""); sys.exit()
+data=d.get("data") or []
+print((data[0].get("max_model_len") if data else "") or "")' 2>/dev/null)"
```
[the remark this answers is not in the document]
MIR-AMD: Resolved — false positive for the active parser: the `python3 -c '...'` lines are at **column 0**, so the Python is valid (same construct as the live hooks, which pass `resolve_ctx_offline.sh`). Minor note: the *commented-out* `/get_server_info` fallback example just below is prefixed `# ` with indentation, so if you enable it, un-comment its Python lines back to column 0 to keep them valid.

scripts/common/agentx/HOW_IT_WORKS.md
```diff
@@ -0,0 +1,358 @@
+# How AgentX works (and why it is accurate)
```
Cemberk, line 1: 1772 of the 5533 lines in this PR are documentation — HOW_IT_WORKS.md (358),
README.md (454), SCENARIOS.md (335), profileREADME.md
(110), plus 334 across the two launcher READme config
schema is explained in at least three of them.

The content is good — §2 here is the clearest statement anywhere of *why* a synthetic
corpus measures a real engine faithfully, an I'd
consolidate to README (schema + env table) + HOW_IT_WORKS (mechanism), fold SCENARIOS
into README as an examples section, and drop templates/README along with the template.

scripts/common/agentx/tests/_stub_server.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
```
Cemberk, line 1: Semantically identical to `scripts/vllm_dissag/tests/_stub_server.py` — both serve a
`/v1/models` card with `max_model_len: 131072` plus an `empty` mode. Not copy-paste,
though: this one factors out a `_send()` helper and handles `/get_server_info`; the
vLLM one inlines the response and reads `PORT` at module scope. Two independent
implementations of the same 40-line stub.

Same story for `resolve_ctx_offline.sh` in each directory — identical assertions,
different port-picking (`$((18000 + RANDOM))` here vs a `socket.bind(0)` helper
there) and different variable naming throughout.

One `scripts/common/agentx/tests/` copy of each, with a `--backend` flag, covers both.

scripts/common/agentx/templates/benchmark_agentic.template.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
```
Cemberk: This template has already drifted from both files it templates. Compare the three
`agentx:BEGIN/END` regions:

    sglang_disagg/benchmark_agentic.sh:43-90  — loops two endpoints
                                                (/v1/models then /get_server_info),
                                                heredoc'd Python
    vllm_dissag/benchmark_agentic.sh:45-78    — single endpoint, inline Python,
                                                different error text
    this template:50-104                      — a third variant

Hand-synced copy markers are the mechanism that produced the drift, so shipping a
template to manage it makes the problem permanent rather than solving it.

The genuine backend-specific surface is three values:

    | | sglang | vllm |
    |---|---|---|
    | port default    | 2322             | BENCHMARK_PORT+1 |
    | mount env prefix| MOONCAKE_*       | NIXL_*           |
    | ctx fallback    | /get_server_info | —                |

Suggest one `scripts/common/benchmark_agentic.sh` taking `--backend sglang|vllm` (or
reading `AGENTIC_BACKEND`), with those three as a `case`. That deletes this template
and both hooks — ~340 lines down to ~150 — and makes drift structurally impossible.

Also worth noting: SGLang and vLLM are the only disagg launchers in this repo.
Building a copy-paste onboarding path for a hypothetical backend #3 is speculative,
and if #3 does arrive a shared script serves it better than a template does.

scripts/sglang_disagg/run_xPyD_models.slurm
```diff
@@ -114,6 +114,14 @@ DP_MODE="${DP_MODE:-0}" #-> mori_io_ep: 1=dp flags + --dp-size/--ep-size; 0=tp f
 SKIP_BENCHMARK="${SKIP_BENCHMARK:-0}"
 SKIP_CURL_TEST="${SKIP_CURL_TEST:-0}"
 
+# Agentic replay shorthand: AGENTIC=1 selects the agentic benchmark hook. The
+# in-container launcher (sglang_disagg_mori_io_ep.sh) then auto-enables server
```
Cemberk: This shorthand sets `BENCHMARK_SCRIPT=agentic` — the exact variable the case statement
three lines below already reads. The repo's established idiom is right here:

    BENCHMARK_SCRIPT=<mode>  ->  case  ->  BENCHMARK_SCRIPT_FILE=<file>

with validation at :443 and a clear error listing valid modes. Adding `agentic)` to
that case at :442 was the correct and sufficient change. `AGENTIC=1` is a second way
to say the same thing, and it's the one the READMEs teach.

Three things follow from using the existing idiom properly:

1. **Drop `AGENTIC=1` (:431-437 here, :118-123 in the SGLang launcher).**
   The repo already has a named-mode benchmark selector (`BENCHMARK_SCRIPT=sweep|long_context`,
  resolved centrally to a filename with validation) in the vLLM launcher. This PR correctly
  extends it with `agentic`, then adds a redundant `AGENTIC=1` shorthand for it, skips the
  selector entirely on the SGLang side, and gates deep code on the resolved *filename*
  rather than the mode name. Using the existing idiom — plus `BENCHMARK_SCRIPT` in
  `models.json` env_vars — removes the shorthand, brings the two launchers back into
  parity, and keeps the filename confined to the one case statement that owns it.

2. **Give SGLang the same selector.** `scripts/sglang_disagg/run_xPyD_models.slurm` has
   no `BENCHMARK_SCRIPT` case at all — it goes straight from `AGENTIC=1` to a raw
   `BENCHMARK_SCRIPT_FILE`, unvalidated. The two launchers should not diverge on how a
   benchmark is chosen. Porting the case statement is ~8 lines and makes SGLang's
   invalid-mode failure as legible as vLLM's.

3. **Gate on the mode, not the filename.** All 8 gate sites test
   `BENCHMARK_SCRIPT_FILE == "benchmark_agentic.sh"` — the resolved implementation
   detail — rather than `BENCHMARK_SCRIPT == "agentic"`. That's what pushes a filename
   down into `connectors/moriio.sh` and the worker-launch path, where renaming a script
   silently changes serving config. The whole point of the two-layer design is that
   only the case statement knows the filename.

scripts/sglang_disagg/run_xPyD_models.slurm
```diff
@@ -300,14 +308,34 @@ echo ""
 USER_NAME=$(whoami)
 MASTER_NODE=$(echo "$SELECTED_NODES" | head -n 1)
 MASTER_ADDR=$(srun --nodes=1 --ntasks=1 --time=00:20:00 --nodelist="$MASTER_NODE" bash -c 'hostname -I')
-MASTER_ADDR=$(echo "$MASTER_ADDR" | awk 'NR==1 {print $1}')
+# Prefer the 10.158.x fabric IP: nodes are multi-homed and picking the first
+# hostname -I address can land on a different subnet, which hangs the socket
+# barrier on mixed-subnet coordination. Cluster-specific (OCI useocpm2m fabric).
+# Fall back to the first address if no 10.158.x is present; fail fast if empty.
+MASTER_ADDR_RAW="$MASTER_ADDR"
```
Cemberk, line 328: Hardcoding a `10.158.x` fabric-subnet preference for MASTER_ADDR/IP resolution is an
OCI cluster fix with no connection to agentic benchmarking, and it changes node
coordination for every existing sweep. It also embeds one cluster's addressing in a
repo meant to run elsewhere — should that be `FABRIC_SUBNET_PREFIX` with the current
value as the default?

scripts/sglang_disagg/run_xPyD_models.slurm
```diff
@@ -398,6 +426,30 @@ docker run --rm \
...
+    -e MAX_MODEL_LEN="${MAX_MODEL_LEN:-}" \
+    -e AGENTIC_CONFIG="${AGENTIC_CONFIG:-}" \
+    -e AGENTIC_WORKLOAD="${AGENTIC_WORKLOAD:-}" \
+    -e SUITE_CORPUS_DIR="${SUITE_CORPUS_DIR:-}" \
+    -e AIPERF_PIN="${AIPERF_PIN:-}" \
+    -e AGENTIC_UTILS_PIN="${AGENTIC_UTILS_PIN:-}" \
+    -e SUITE_CORPUS_FORCE="${SUITE_CORPUS_FORCE:-}" \
+    -e AGENTIC_STRICT_CONTEXT="${AGENTIC_STRICT_CONTEXT:-}" \
+    -e AGENTIC_HF_ISL_TAIL="${AGENTIC_HF_ISL_TAIL:-}" \
+    -e HF_HOME="${HF_HOME:-}" \
```
Cemberk: `-e HF_HOME="${HF_HOME:-}"` is unconditional, so it applies to the default
non-agentic sweep too. This is the only place HF_HOME enters this container and
`sglang_disagg_inference.ubuntu.amd.Dockerfile` doesn't set it, so when the host has
HF_HOME unset the container now gets `HF_HOME=""` where it previously got nothing.
`huggingface_hub` reads that as present-but-empty, making `HF_HUB_CACHE` resolve to
a relative `hub/` under CWD.

I haven't run this to confirm the downstream effect, but it does contradict the
"non-agentic sweeps are byte-for-byte unchanged" claim, and the `:+` fix above makes
it moot either way.

scripts/sglang_disagg/run_xPyD_models.slurm
```diff
@@ -398,6 +426,30 @@ docker run --rm \
     -e BENCHMARK_COMBINATIONS="${BENCHMARK_COMBINATIONS:-1024/1024 8192/1024}" \
     -e DOCKER_IMAGE_NAME=${DOCKER_IMAGE_NAME:-unknown} \
     -e USE_CX7_NICS=${USE_CX7_NICS:-0} \
+    -e BENCHMARK_SCRIPT_FILE="${BENCHMARK_SCRIPT_FILE:-}" \
```
Cemberk: 24 enumerated `-e` lines, using `-e VAR="${VAR:-}"` — which sets the variable to an
*empty string* in the container when unset on the host, rather than leaving it unset.

The vLLM sibling launcher already uses the correct idiom for exactly this
(`scripts/vllm_dissag/run_xPyD_models.slurm:618-640`):

    ${VAR:+-e VAR=$VAR} \

Please switch to that form. Better still, loop over the prefix so new knobs don't
require editing two launchers:

    for v in $(compgen -v AGENTIC_); do docker_env+=(-e "$v=${!v}"); done

scripts/sglang_disagg/run_xPyD_models.slurm
```diff
@@ -114,6 +114,14 @@ DP_MODE="${DP_MODE:-0}" #-> mori_io_ep: 1=dp flags + --dp-size/--ep-size; 0=tp f
 SKIP_BENCHMARK="${SKIP_BENCHMARK:-0}"
 SKIP_CURL_TEST="${SKIP_CURL_TEST:-0}"
 
+# Agentic replay shorthand: AGENTIC=1 selects the agentic benchmark hook. The
+# in-container launcher (sglang_disagg_mori_io_ep.sh) then auto-enables server
+# metrics + radix prefix cache for BENCHMARK_SCRIPT_FILE=benchmark_agentic.sh.
+if [[ "${AGENTIC:-0}" == "1" ]]; then
```
Cemberk: This shorthand is redundant with madengine. `models.json` already sets arbitrary env
for a run, and this file already reads `BENCHMARK_SCRIPT_FILE` from env and forwards
it into the container — so a `models.json` entry with

    "env_vars": { "BENCHMARK_SCRIPT_FILE": "benchmark_agentic.sh",
                  "AGENTIC_WORKLOAD": "small" }

reaches the identical code path with zero launcher changes. This block gives us a
second way to set one variable in a repo whose entry point is `madengine run --tags`.

Suggest dropping it in favour of the entries described in the review body. Same
applies to `scripts/vllm_dissag/run_xPyD_models.slurm:434`.

scripts/sglang_disagg/models.yaml
```diff
@@ -9,7 +9,15 @@
...
 # - prefill/decode.<tp|dp>: role + mode specific flags
-# - experimental_flags: optional extra CLI flags (omit if empty)
+# - experimental_flags: optional extra CLI flags appended after role/mode flags on BOTH prefill and decode workers.
+#   Use for try-outs that are independent of TP vs DP (PARALLEL_MODE); omit or "" when unused.
+#
+# Radix prefix cache: base_flags below carry "--disable-radix-cache" for the default
+# xPyD performance sweeps (clean, cache-free throughput). The agentic replay benchmark
+# needs the cache ON, so the launcher (sglang_disagg_mori_io_ep.sh) STRIPS
+# "--disable-radix-cache" from PREFILL/DECODE configs when
+# BENCHMARK_SCRIPT_FILE=benchmark_agentic.sh (or ENABLE_RADIX_CACHE=1). Do not remove it
```
Cemberk: This comment correctly identifies the hazard — "Do not remove it here, or the perf
sweeps lose their cache-free baseline" — but a warning comment is standing in for
separation. `sglang_disagg_mori_io_ep.sh:195` string-strips `--disable-radix-cache`
from whichever entry is selected, so all 10 entries carrying that flag now have two
serving configurations sharing one definition, distinguished only by an env var set
three layers up.

The same pattern is at `scripts/vllm_dissag/models.yaml:35`, where the note ends with
"there is nothing to set here. Do not add a prefix-cache flag to these entries" —
i.e. the config that governs the agentic run is deliberately *not* in the file that
describes the model.

If the agentic runs were their own `models.json` entries, this would be ordinary
declared config instead of a documented landmine.

Cemberk (changes requested):
The corpus design is sound and I'd like this in. Fixed offered load — token counts
plus `hash_ids` block structure, seeded and reproducible — with achieved metrics
measured per engine is the right way to make SGLang and vLLM comparable, and the
growing-shared-prefix shape is something the existing `1024/1024` sweep genuinely
cannot produce. `scripts/common/agentx/` (config resolution, generator, verifier,
filter) is good, well-commented work.

My concerns are about integration shape rather than the benchmark itself.

**The headline one: the gate lives in the shared launcher rather than in a model
definition, so every existing model entry silently acquires a second serving
configuration.**

    scripts/sglang_disagg/models.json  — 16 entries
    scripts/vllm_dissag/models.json    — 12 entries
28 declared workloads, 56 reachable ones. The undeclared 28 differ from their twins
in prefix-cache state, benchmark driver, and metrics flags. Three were validated
(Qwen3-32B, DeepSeek-V3, DeepSeek-V3-5layer); the other 25 inherit an untested mode
by construction. Three consequences:

1. *Shared per-model config is mutated in place.* The launcher string-strips
   `--disable-radix-cache` from whichever `models.yaml` entry is selected (10 carry
   it). See the inline note on `models.yaml`.

2. *Results collide in the dashboard.* `scripts/sglang_disagg/parse_to_csv.py:159`
   keys `perf.csv` on the `models.json` entry name. Both modes run under the same
   name, so a cache-free sweep row and a cache-on agentic replay row are
   indistinguishable downstream — same key, different serving config, incomparable
   metrics. For a repo whose stated purpose is performance tracking and dashboards,
   this is the part I'd most want resolved before merge.

3. *The mode isn't discoverable.* Nothing in `madengine run --tags` surfaces it, so
   it can't be scheduled, can't regress in CI, and only exists for someone who read
   the README and knows which variable to export.

#197 (61b5301) moved models to per-directory `models.json` immediately before this
branch, which makes the alternative cheap: declare the agentic runs as their own
entries in `scripts/sglang_disagg/models.json` and `scripts/vllm_dissag/models.json`,
tagged `agentic`. That makes the 2N configurations explicit, lets you declare only
the ones actually validated rather than all 28, gives each a distinct dashboard key,
and makes both the `AGENTIC=1` shorthand and most of the 8 filename gate sites
unnecessary — the entry carries the config.

The launcher hook mechanism itself is right. Reusing the xPyD topology, MoRI setup,
and node coordination is exactly what should happen; duplicating them would be far
worse. It's the *implicit* mode-per-existing-entry I'd push back on, not the reuse.

Secondary concerns, detailed inline:
- ~330 lines of shell/Python duplicated across the two launchers, and the template
  added to manage that duplication has already drifted from both copies it templates.
- The agentic path is selected by comparing a filename string in 8 places across
  shared production code.
- One env-forwarding block uses `-e VAR=""` where the sibling launcher already uses
  the correct `${VAR:+-e VAR=$VAR}` idiom; `HF_HOME` is affected on the default path.
- Three unrelated fixes are bundled in (fabric IP resolution, barrier port, benchmark
  exit-status propagation). The exit-status one is a real bug fix that should land —
  just on its own.
- 1772 of 5533 lines are docs with meaningful overlap between them.

Happy to re-review quickly once the shape is settled.

MIR-AMD:
## Follow-up: parser + driver failure-masking fixes

Pushed two commits to `miali/agentic-suite`:
- `8f40a3a` fix(agentx): resolve N1 parser + robustness guards
- `763c435` fix(agentx): resolve shell driver failure masking

### Bugs resolved
- **N1 — silent data loss (HIGH):** the fallback YAML parser dropped block sequences indented at the *same* level as their key, so `workloads:` followed by a same-indent `-` list resolved to zero workloads with no error. Fixed by descending into same-indent block sequences.
- **Suite masked failures (MED):** the agentic driver ran each workload with `|| true`, so the suite exited 0 even when every workload failed. Replaced with `_suite_failed` propagation to the suite exit code (RUN_INVALID marker preserved).
- **Launcher masked failures (MED):** the sglang in-container `... | tee` pipe returned tee's (always-0) exit. Added `set -o pipefail`.
- **AGENTIC_CONFIG `$HOME` not expanded (MED):** launcher forwarded a literal `$HOME/...` into the container (where `$HOME=/root`) -> "file not found". Now expanded on the submit host before the docker `-e` forward.
- **Robustness guards:** gen (`block_size >= 1`; turns non-empty / equal-length / positive weights), verify (session JSON missing `requests`), config (preset-not-found) now fail with actionable messages instead of raw tracebacks.

### Testing
- Offline gate `scripts/common/agentx/tests/run_offline.sh`: **35 -> 43 passed, 0 failed** (+8 regression assertions covering every fix above, including comment-excluding static guards for the shell fixes).
- N1 and the driver defects were surfaced by an overnight offline fuzz campaign and a reduced single-concurrency agentic full-suite re-run on DeepSeek-V3 (vLLM MoRI-EP), which completed with a clean exit-0 aggregate after the fixes.

### Deferred (separate efforts)
- N2 (top-level flow) and N3 (PyYAML 1.1 scalar-type parity) fallback-parser gaps.

Cemberk (changes requested):
Re-reviewed after the updates. The shape concerns from my last pass are genuinely resolved — thanks for working through them:

- **Explicit `models.json` entries**, `agentic`-tagged, declaring only the four validated configurations rather than doubling all 28. This was my headline concern and it's the right fix: distinct dashboard keys, schedulable via tags, no undeclared second mode per existing entry.
- **Mode-based gating** (`BENCHMARK_SCRIPT == agentic`) replaced the filename comparison — no filename gates remain; the surviving `BENCHMARK_SCRIPT_FILE` references are comments and the case-mapping.
- **~330 duplicated lines collapsed** into one `--backend` script behind 12-line per-backend shims, which also retires the drifting template.
- **`${VAR:+-e VAR=$VAR}`** applied consistently, including `HF_HOME`.
- **Docs down to 32%** of the diff, launcher READMEs trimmed to 17/15 `AGENTIC_` mentions against 58 in the canonical one.

Offline gate is 43/43 on my checkout. Two blockers below, plus a layout point, then smaller items.

---

## 1. Blocking — the two vLLM agentic entries cannot launch

`scripts/sglang_disagg/run_xPyD_models.slurm:406` expands `$HOME` before the docker forward:

```bash
[ -n "${AGENTIC_CONFIG:-}" ] && AGENTIC_CONFIG="$(eval echo "$AGENTIC_CONFIG")"
```

The vLLM launcher has no equivalent — it only forwards, at `scripts/vllm_dissag/run_xPyD_models.slurm:581`. Both vLLM agentic entries set:

```json
"AGENTIC_CONFIG": "$HOME/MAD/scripts/common/agentx/agentic.example.yaml"
```

so a literal `$HOME` reaches the container, where `$HOME=/root` → file not found. This is the same defect `8f40a3a` fixed, applied to only one of the two launchers.

It survives the gate because `scripts/common/agentx/tests/run_offline.sh:309` asserts the guard *only* against the sglang launcher. Fix is those two lines in the vLLM launcher, plus extending that assertion to cover both.

While there: `eval echo` on a path executes any command substitution present in the value. `AGENTIC_CONFIG="${AGENTIC_CONFIG/#\$HOME/$HOME}"` does the same job without `eval`.

## 2. Blocking — declared entries point at the example config

Three of the four entries use `agentic.example.yaml`, whose own header reads *"Copy this to agentic.yaml (or point AGENTIC_CONFIG at it) and edit."* It carries seven demonstration workloads. Dry-run against DeepSeek-V3's discovered 163840 window:

- **19 measured windows × 900s = 4.75h** of measured time alone, before corpus generation, HF download, and warmup.
- Every workload lands `WARN` and has its context capped.
- `my_corpus` points at `input_dir: /tmp/agentx_corpora/conformance_256k`. `scripts/common/benchmark_agentic_suite.sh:118-119` calls `agentic_die` → `exit 1` when that directory is missing. It is workload 6 of 7, so a real run spends ~3.25h across 13 windows and then hard-aborts.
- `suite_summary.json` is written *after* the loop, so that abort yields **no summary at all** — a multi-hour job reports nothing downstream.
- `my_case` is a `GLM-5.2-MXFP4` placeholder.

The `DRY_RUN` path skips the `input_dir` check (`:117`), which is why the offline gate cannot catch this.

Suggested fix: point the entries at a real committed config (`agentx/configs/nightly.yaml` or similar) containing only validated workloads, or follow the Qwen entry's lead and use `AGENTIC_WORKLOAD=<preset>`. Keep `agentic.example.yaml` as documentation, and comment out `my_corpus` / `my_case` there so a copy-paste user doesn't hit the same abort.

Worth fixing independently of that: a missing `input_dir` aborting the entire suite is inconsistent with how the driver handles every other failure — replay failures set `_suite_failed=1` and continue (`:192-194`), context misfits `continue` (`:126-129`). A missing input dir should mark that workload failed and carry on.

## 3. Layout — six agentic-only files sit outside `agentx/`

`scripts/common/` held exactly one file on develop (`rocm_paths.sh`, genuinely cross-recipe). This PR makes it seven, six of them agentic-only:

```
agentic_lib.sh (491)  benchmark_agentic.sh (152)  benchmark_agentic_suite.sh (258)
preflight_weights.sh (48)  validate_agentic_result.sh (75)  verify_agentic_image.sh (133)
```

Everything else in the suite already lives in `agentx/`, and nothing outside the agentic path references any of the six. Moving all six into `scripts/common/agentx/` keeps `common/` as the shared-across-recipes namespace it was.

The move is cheap, since the shims already resolve through a candidate list plus an `AGENTIC_LIB` override. Touch points:

- `agentic_lib.sh:49` — `AGENTX_DIR` is built as `$(dirname)/agentx`; becomes `$(dirname "${BASH_SOURCE[0]}")`
- locator candidate lists in `benchmark_agentic.sh:43-49` and `benchmark_agentic_suite.sh:19`
- the two per-backend shims
- `agentx/tests/run_offline.sh:20,41`
- README path references

Two of the six deserve to go a step further: `preflight_weights.sh` and `verify_agentic_image.sh` have **no callers anywhere** — they're manual operator utilities, and `preflight_weights.sh` hardcodes OCI node numbers (`008 030 038 083 087...`) in an error string. Those fit `agentx/tests/` alongside `harvest.sh` and `submit_matrix.sh` rather than `common/`.

## 4. Still bundled — the three unrelated fixes

The fabric IP resolution, `BARRIER_PORT` plumbing, and `set -o pipefail` propagation are unrelated to agentic and affect every non-agentic sweep. `pipefail` in particular is a real fix that could land on its own today, ahead of this.

If they stay: `BARRIER_PORT` at `scripts/vllm_dissag/run_xPyD_models.slurm:490` is used inside the single-quoted `srun bash -c '...'` but is never exported in that script, so it resolves only when the caller happened to export it.

## 5. Minor

- Exec bits are inconsistent: `benchmark_agentic_suite.sh` is 644 while `benchmark_agentic.sh` is 755; within `agentx/tests/` only `resolve_ctx_offline.sh` is 755.
- The suppressed Copilot nits are still open and all fair: `harvest.sh` always exits 0 despite tracking `overall_fail`; the shim's `/health` returns body `"ok"` alongside HTTP 503; `verify_agentx_profile.py:83` uses a list comprehension for its side effect.

---

The two blockers are small in code but each produces a failed or empty run on the declared entries, so I'd want those plus the `agentx/` move before merge. Everything else here is optional or splittable. Happy to re-review quickly.

Cemberk (dismissed):
LGTM 👍

PR #216, opened 2026-08-19
MLPerf Llama-3.1-8B training and inference on ROCm/SLURM (reopen of #208)

mkuznet1:
_Recreation of #208, rebased onto current `mad-rccl`._

Adds two MLPerf Llama-3.1-8B workloads to MAD — reference training (MLCommons `small_llm_pretraining/nemo`) and reference inference (loadgen + vLLM SUT) — brought up on gfx950 and driven through madengine on SLURM.

## Training — `pyt_mlperf_training_llama-3.1-8b`

- `docker/pyt_mlperf_training.ubuntu.amd.Dockerfile` — NeMo 2.7.3 / Megatron-LM / TransformerEngine on a **py3.12** ROCm base (NeMo 2.7.3 cannot be installed on py3.13), optionally building rdma-core 63.0 from source for the Broadcom Thor2 `bnxt_re` provider.
- `scripts/pyt_mlperf_training/run.sh` — checks out the pinned `mlcommons/training` ref in-container, fetches the tokenizer and preprocessed C4 on rank 0 behind sentinels, then launches torchrun.
- `scripts/pyt_mlperf_training/mlperf_pretrain_entrypoint.py` — wraps the reference recipe: GBS/DP/MBS divisibility gate, rank-0 data-index build published through a TCPStore, MLLOG compliance keys.
- `scripts/pyt_mlperf_training/extract_perf.py` — step-time percentiles, cluster and per-GPU throughput, and MFU against `MLPERF_PEAK_BF16_TFLOPS`. The peak itself is written to the CSV so the divisor can be audited after the fact.

## Inference — `pyt_mlperf_inference_llama-3.1-8b`

- `docker/pyt_mlperf_inference.ubuntu.amd.Dockerfile` — vLLM ROCm base plus the accuracy dependencies the reference harness imports.
- `scripts/pyt_mlperf_inference/run.sh` — Offline scenario, performance and/or accuracy, ROUGE evaluated against the targets selected by the loadgen model name. Single node by design: the reference harness is one process driving one vLLM instance and has no multi-node mode.

## `mad-slurm-multinode` skill

Four manifest templates (a perf and a smoke variant per workload), plus a `validate_manifest.sh` relaxation so single-node manifests are no longer required to carry inter-node RDMA transport vars — with `nodelist` cardinality counted as a node-count source, so a `nodelist`-only multi-node manifest still gets the checks. Reference notes cover the cluster archetype, the four data mounts, and the failure modes hit during bring-up.

## Verification

| run | result |
| --- | --- |
| training, 1 node, smoke | pass |
| training, 2 nodes, 100 steps | pass — 1.408 s/step, 372 238 cluster tokens/s, `mfu_pct` 44.84 against `peak_bf16_tflops` 2500 |
| inference, 1 node, Offline, edge 5000 samples, TP=8 | `Result is : VALID` — 38.10 samples/s, 4877 tokens/s, rouge1 39.07, `accuracy_pass=1` |

Both images were rebuilt from scratch after the review fixes, so the new build-time gates (the `nltk.download` return-value check and the `bnxt_re` provider presence check) are exercised rather than assumed.

## Known gaps

- **4 nodes hang.** Ranks 8-23 sit in `WorkNCCL(SeqNum=27, OpType=ALLREDUCE, NumelIn=1)` and abort at the 600 s default process-group timeout, while the other 16 ranks report `last enqueued NCCL work: 26` and never reach that collective. NCCL topology is fine (all rings and trees over `NET/IB`, GDR on all 8 `bnxt_re` HCAs) and 2 nodes pass, so this is a scale threshold rather than a broken transport. Not root-caused yet.
- `MEGATRON_COMMIT` and `TE_COMMIT` pin branches rather than SHAs. Flipping them to SHAs needs a rebuild plus a re-verify run to mean anything, so it is deliberately left as follow-up.
- The training image has no `ARG RCCL_*`, unlike the sglang overlay, so it cannot pin an RCCL candidate for validation on this branch. Feature gap, tracked separately.


docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,392 @@
...
+ARG BASE_DOCKER
+
+ARG MLPERF_WORKSPACE=/workspace
+ARG BUILD_GPU_TARGETS=gfx950
+ARG MEGATRON_REPO=https://github.com/ROCm/Megatron-LM.git
+# NeMo 2.7.3 pins megatron-core 0.15.0rc8 (requirements/manifest.json ->
+# NVIDIA/Megatron-LM@bf1a5035). Anything newer breaks the recipe at runtime:
+# 0.16 dropped `no_weight_decay_cond` from get_megatron_optimizer() and the
+# dist_checkpointing tensorstore strategy NeMo imports.
+ARG MEGATRON_COMMIT=core_r0.15.0_rocm
```
i-kosarev, line 19: `MEGATRON_COMMIT` pins a branch, not a SHA — reproducibility drifts silently across rebuilds.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,392 @@
...
+ARG NEMO_REPO=https://github.com/NVIDIA/NeMo.git
+ARG NEMO_COMMIT=v2.7.3
+ARG NEMO_RUN_REF=v0.10.0
+ARG TE_REPO=https://github.com/ROCm/TransformerEngine.git
+ARG TE_COMMIT=release_v2.17_rocm
```
i-kosarev, line 24: Same issue — `TE_COMMIT` is a branch (`release_v2.17_rocm`), not a SHA.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,392 @@
...
+ARG MLPERF_BUILD_PROFILE=full
+ARG TE_WHEEL_URL=
+ARG TE_ROCM_WHEEL_URL=
+# Optional rdma-core from source (63.0 carries the Broadcom Thor2 bnxt_re
+# ibv_create_qp EFAULT fix). Empty string keeps the distro rdma-core.
+ARG RDMA_CORE_VERSION=63.0
+
+ENV WORKSPACE_DIR=${MLPERF_WORKSPACE}
+ENV DEPS_DIR=${MLPERF_WORKSPACE}/deps
```
i-kosarev, line 33: No `RCCL_REPO/BRANCH/COMMIT` build-args here, unlike PR #388/#389 — this image can't pin a candidate RCCL commit for validation.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,392 @@
...
+
+RUN git clone --recursive "${MEGATRON_REPO}" megatron_lm && \
+    cd megatron_lm && \
+    git checkout "${MEGATRON_COMMIT}" && \
+    pip uninstall -y megatron-core || true && \
+    pip install -e . && \
+    cd megatron/core/datasets && \
+    make
+
+ENV PYTHONPATH="${DEPS_DIR}/megatron_lm"
```
i-kosarev, line 100: This overwrites `PYTHONPATH` instead of appending (`:${PYTHONPATH}`), dropping whatever the base image set.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,392 @@
...
+      fi; \
+      ninja install && ldconfig; \
+      cd / && rm -rf /tmp/rdma-core; \
+    else \
+      echo "RDMA_CORE_VERSION empty — keeping distro rdma-core"; \
+    fi
+
+RUN if [[ -n "${RDMA_CORE_VERSION}" ]]; then \
+      readelf -d /usr/lib/x86_64-linux-gnu/libibverbs.so.1 | grep SONAME && \
+      { ls /usr/lib/x86_64-linux-gnu/libibverbs/libbnxt_re-rdmav*.so 2>/dev/null | head || true; }; \
```
i-kosarev: `|| true` means the one thing rdma-core 63.0 was built to fix (bnxt_re provider present) can silently fail here.

docker/pyt_mlperf_inference.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+print("patched", path)
+PY
+
+RUN pip install "${MLPERF_INFERENCE_DIR}/loadgen" && \
+    python3 -c "import mlperf_loadgen; print('loadgen', mlperf_loadgen.__file__)"
+
+# The harness `requirements.txt` pins `vllm==0.6.3` and `transformers==4.46.2`;
+# installing it would replace the ROCm vLLM build that this base image exists
+# for. Only the imports the base image is actually missing are added here.
+RUN pip install nltk rouge-score absl-py
```
i-kosarev: `evaluate` is never installed but is imported below (L63/L66) — likely build-breaking.

docker/pyt_mlperf_inference.ubuntu.amd.Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+
+# Pre-seed what the accuracy pass would otherwise fetch mid-run: nltk's sentence
+# tokenizer and the `rouge` metric module from the Hub. HF_MODULES_CACHE is set
+# explicitly so a run-time HF_HOME override cannot invalidate the warm cache.
+RUN python3 -c "import nltk; nltk.download('punkt', download_dir='${NLTK_DATA}'); nltk.download('punkt_tab', download_dir='${NLTK_DATA}')" && \
```
i-kosarev: `nltk.download()` returns `False` on failure instead of raising — a transient network error bakes an empty cache as "success".

scripts/pyt_mlperf_training/run.sh
```diff
@@ -0,0 +1,526 @@
...
+      echo "[data-fetch] rank=0 tokenizer already present at ${TOKENIZER_PATH}"
+    else
+      local hf_token="${MAD_SECRETS_HFTOKEN:-${HF_TOKEN:-}}"
+      if [[ -z "${hf_token}" ]]; then
+        echo "[data-fetch] ERROR: tokenizer missing at ${TOKENIZER_PATH} and no HF token provided (MAD_SECRETS_HFTOKEN/HF_TOKEN)."
+        return 1
+      fi
+      echo "[data-fetch] rank=0 downloading meta-llama/Llama-3.1-8B tokenizer -> ${TOKENIZER_PATH}"
+      rm -f "${tokenizer_sentinel}"
+      HF_TOKEN="${hf_token}" huggingface-cli download \
```
i-kosarev, line 168: `set -x` is still on here — the HF token value gets printed to the trace log.

scripts/pyt_mlperf_training/run.sh
```diff
@@ -0,0 +1,526 @@
...
+      echo "[data-fetch] rank=0 preprocessed C4 already present at ${PREPROCESSED_PATH}"
+    else
+      local dataset_uri="${MLPERF_DATASET_URI:-https://training.mlcommons-storage.org/metadata/llama-3-1-8b-preprocessed-c4-dataset.uri}"
+      local downloader_url="${MLPERF_R2_DOWNLOADER:-https://raw.githubusercontent.com/mlcommons/r2-downloader/refs/heads/main/mlc-r2-downloader.sh}"
+      echo "[data-fetch] rank=0 downloading MLCommons preprocessed C4 dataset -> ${PREPROCESSED_PATH}"
+      echo "[data-fetch] URI=${dataset_uri}"
+      rm -f "${dataset_sentinel}"
+      (
+        cd "${PREPROCESSED_PATH}"
+        bash <(curl -fsSL "${downloader_url}") -d "${PREPROCESSED_PATH}" "${dataset_uri}"
```
i-kosarev: `curl` failure inside `bash <(curl ...)` isn't caught by `set -e`, so a bad download URL still lets rank 0 publish the "ready" sentinel.

scripts/pyt_mlperf_training/extract_perf.py
```diff
@@ -0,0 +1,609 @@
...
+        rows.append((str(gpus_per_node), "gpus_per_node"))
+    if result.get("global_batch_size") is not None:
+        rows.append((str(result["global_batch_size"]), "global_batch_size"))
+    if result.get("n_gpus") is not None:
+        rows.append((str(result["n_gpus"]), "total_gpus"))
+
+    # Keep the legacy status marker so MAD's existing ``torchrun_launch_success``
+    # detection still sees a green signal even when the perf numbers are the
+    # primary value. Also include the submittable/status flag from MLLOG.
+    status_marker = "torchrun_launch_success" if result.get("submittable") is not None else "launch_success"
```
i-kosarev: `submittable` is always a bool, so the else-branch is dead — a failed run still gets tagged `torchrun_launch_success`.

scripts/pyt_mlperf_training/extract_perf.py
```diff
@@ -0,0 +1,609 @@
...
+# standard MLPerf/megatron FLOPs-per-token formula (fwd+bwd, no recompute).
+PARAM_COUNT: dict[str, int] = {
+    "8b": 8_030_000_000,
+    "70b": 70_553_000_000,
+    "405b": 405_853_000_000,
+}
+
+# MI325X dense BF16 matrix peak (TFLOP/s). Override via --peak-bf16-tflops for
+# other accelerators.
+DEFAULT_PEAK_BF16_TFLOPS = 1307.0
```
i-kosarev, line 60: If `MLPERF_PEAK_BF16_TFLOPS` isn't set, `mfu_pct` silently uses the wrong accelerator's peak, and the peak used is never written to the CSV for later audit.

scripts/pyt_mlperf_training/mlperf_pretrain_entrypoint.py
```diff
@@ -0,0 +1,411 @@
...
+        seq_length=seq_length,
+        tokenizer_path=args.tokenizer_path,
+        seed=args.seed,
+        use_full_dataset=args.use_full_dataset,
+    )
+
+    eval_every_n_batches = math.ceil(args.eval_every / args.gbs)
+    eval_batches = math.ceil(args.eval_tokens / args.gbs)
+    start_eval_at = (
+        math.ceil(args.start_eval_at / args.gbs) if args.start_eval_at == 0 else eval_every_n_batches
```
i-kosarev: This condition is inverted — any nonzero `start_eval_at` gets discarded and replaced with `eval_every_n_batches`.

.claude/skills/mad-slurm-multinode/scripts/validate_manifest.sh
```diff
@@ -64,16 +64,21 @@ denv  = dep.get("env_vars") or {}
 def is_ph(v):
     return isinstance(v, str) and ("<FILL" in v or v.strip() == "")
 
-# 3) NCCL_IB_HCA present + non-empty in both env blocks, and equal
+# 3) NCCL_IB_HCA present + non-empty in both env blocks, and equal. Single-node
+# runs carry no inter-node traffic, so they are allowed to omit the RDMA vars.
+multi_node = max(slurm.get("nodes") or 1, dist.get("nnodes") or 1) > 1
```
i-kosarev: This ignores `nodelist`-only allocations (no `nodes`/`nnodes` set), so such manifests skip the RDMA/HCA symmetry check that catches multinode hangs.

i-kosarev (commented):
A few things

.claude/skills/mad-slurm-multinode/assets/manifests/mlperf_training_llama-3.1-8b_smoke.template.json
```diff
@@ -0,0 +1,136 @@
...
+    }
+  },
+  "context": {
+    "docker_env_vars": {
+      "MAD_SYSTEM_GPU_ARCHITECTURE": "<FILL_GPU_ARCH  gfx950 = MI355X, gfx942 = MI300X, see references/cluster-types.md>",
+      "NCCL_DEBUG": "INFO",
+      "NCCL_DEBUG_SUBSYS": "INIT,NET",
+      "NCCL_IB_DISABLE": "0",
+      "NCCL_NET": "IB",
+      "NCCL_IB_HCA": "<FILL_NCCL_IB_HCA  data-plane HCA list, see references/cluster-types.md>",
```
i-kosarev: This is left as a `<FILL_...>` placeholder rather than omitted, so this single-node smoke manifest fails its own validator's new single-node relaxation (verified by running it).

mkuznet1:
Multi-node follow-up: 2 nodes pass (16 GPUs, 1.41 s/step, mfu 44.8%). 4 nodes hang: ranks 8-23 sit in WorkNCCL(SeqNum=27, ALLREDUCE, NumelIn=1) and abort at the 600s default PG timeout, while ranks 0-7/24-31 report "last enqueued NCCL work: 26". NCCL topology is fine (all rings/trees over NET/IB, 8x bnxt_re GDR), DP=32 TP=PP=1. Not root-caused.

scripts/pyt_mlperf_training/mlperf_pretrain_entrypoint.py
```diff
@@ -0,0 +1,411 @@
...
+    )
```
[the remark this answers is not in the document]
mkuznet1: Intentional, and now documented at `mlperf_pretrain_entrypoint.py:261`: it mirrors mlcommons/training `llama31/pretrain_llama31.py`, where a non-zero `--start_eval_at` is discarded in favour of `eval_every_n_batches`. Diverging would break MLPerf compliance, so the logic stays and the comment cites the reference. d148cff

scripts/pyt_mlperf_training/mlperf_pretrain_entrypoint.py
```diff
@@ -0,0 +1,411 @@
...
+    # GBS / DP / MBS divisibility check: matches what NeMo's MegatronStrategy
+    # asserts later, but produces a much clearer error before we touch CUDA.
+    dp = world_size  # tp=pp=cp=1 for the 8B recipe
+    if args.gbs % dp != 0:
+        raise RuntimeError(
+            f"GBS={args.gbs} is not divisible by DP={dp} "
+            f"(world_size={world_size}). Pick GBS as a multiple of {dp}."
+        )
+    mini_batch = args.gbs // dp
+    if mini_batch % args.mbs != 0:
```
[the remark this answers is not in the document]
mkuznet1: Removed the ambiguity instead of widening the check: `--tensor_parallel_size` was never fed to the trainer strategy (only read back for the MLLOG DP figure), so anything but 1 is now rejected at `mlperf_pretrain_entrypoint.py:221`. The 8B recipe is tp=pp=cp=1, so DP=world_size holds. d148cff

scripts/primus_megatron-lm/run.sh
```diff
@@ -0,0 +1,200 @@
...
+elif [[ "$MODEL_REPO" == "primus_pyt_megatron_lm_train_qwen-3-30b" ]]; then
+  model="Qwen-3-30B"
+elif [[ "$MODEL_REPO" == "primus_pyt_megatron_lm_train_qwen-3-235b" ]]; then
+  model="Qwen-3-235B"
+fi
+
+# Run primus pytorch setup script
+echo "Running setup script to download tokenizers"
+bash ./primus_megatron-lm_benchmark_setup.sh -m $model
+
```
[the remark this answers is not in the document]
mkuznet1: Out of scope: `scripts/primus_megatron-lm/run.sh` is not touched by this PR, which only adds the MLPerf training/inference flow. Worth a separate issue against `mad-rccl`.

scripts/primus_megatron-lm/run.sh
```diff
@@ -0,0 +1,200 @@
...
+export HF_TOKEN=$MAD_SECRETS_HFTOKEN
+
+# Parse named arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --model_repo) MODEL_REPO="$2"; shift ;;
+        *) echo "Unknown parameter passed: $1"; usage ;;
+    esac
+    shift
+done
```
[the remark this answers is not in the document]
mkuznet1: Out of scope: `scripts/primus_megatron-lm/run.sh` is not touched by this PR, which only adds the MLPerf training/inference flow. Worth a separate issue against `mad-rccl`.

scripts/primus_megatron-lm/primus_megatron-lm_benchmark_setup.sh
```diff
@@ -0,0 +1,184 @@
...
+export HF_HOME=/workspace/huggingface
+
+# Parse named arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        -m) MODEL_NAME="$2"; shift ;;
+        *) echo "Unknown parameter passed: $1"; usage ;;
+    esac
+    shift
+done
```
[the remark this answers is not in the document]
mkuznet1: Out of scope: `scripts/primus_megatron-lm/primus_megatron-lm_benchmark_setup.sh` is not touched by this PR, which only adds the MLPerf training/inference flow. Worth a separate issue against `mad-rccl`.

scripts/sglang_disagg/run.sh
```diff
@@ -0,0 +1,221 @@
...
+export IO_EP_TP_SIZE="${IO_EP_TP_SIZE:-${SGLANG_TP_SIZE:-8}}"
+# madengine forwards MASTER_ADDR (rank-0 host) into the container; use it as the
+# rendezvous address for in-container IP discovery below.
+export MASTER_ADDR="${MASTER_ADDR:-127.0.0.1}"
+
+# IPADDRS = comma-separated node IPs in rank order. It may be supplied
+# explicitly via env/arg or SGLANG_NODE_IPS. madengine's container runner does
+# NOT forward SGLANG_NODE_IPS into the container, so when it is absent we
+# self-discover the node IPs in-container using only the forwarded
+# MASTER_ADDR/NODE_RANK/NNODES, mirroring scripts/vllm_dissag/run.sh.
```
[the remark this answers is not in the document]
mkuznet1: Out of scope: `scripts/sglang_disagg/run.sh` is not touched by this PR, which only adds the MLPerf training/inference flow. Worth a separate issue against `mad-rccl`.

scripts/pyt_mlperf_training/run.sh
```diff
@@ -0,0 +1,526 @@
...
+      )
+    fi
```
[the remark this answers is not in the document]
mkuznet1: The failure mode is fixed, the pinning is not. The downloader is fetched to a temp file, size-checked, and the `.ready` sentinel is gated on C4 shards actually being present (`run.sh:196`, 9723a14). The ref is still `refs/heads/main` by default but overridable via `MLPERF_R2_DOWNLOADER`; pinning a SHA there is a follow-up.

.claude/skills/mad-slurm-multinode/assets/manifests/mlperf_training_llama-3.1-8b.template.json
```diff
@@ -0,0 +1,136 @@
...
+      "registry_image": null,
+      "registry": null,
+      "gpu_vendor": "AMD"
+    }
+  },
+  "built_models": {
+    "rocm-mlperf-training-llama31-8b": {
+      "name": "pyt_mlperf_training_llama-3.1-8b",
+      "url": "https://github.com/mlcommons/training",
+      "dockerfile": "docker/pyt_mlperf_training",
```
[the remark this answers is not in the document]
mkuznet1: False positive: madengine treats `built_models.dockerfile` as a prefix and globs `<value>.*` (`docker_builder.py:587`), matching the root `models.json` convention (`"dockerfile": "docker/primus"`). The full path is on line 6, in `built_images.dockerfile`. Verified by a real build + run of both images.

.claude/skills/mad-slurm-multinode/assets/manifests/mlperf_training_llama-3.1-8b_smoke.template.json
```diff
@@ -0,0 +1,136 @@
...
+      "registry_image": null,
+      "registry": null,
+      "gpu_vendor": "AMD"
+    }
+  },
+  "built_models": {
+    "rocm-mlperf-training-llama31-8b": {
+      "name": "pyt_mlperf_training_llama-3.1-8b",
+      "url": "https://github.com/mlcommons/training",
+      "dockerfile": "docker/pyt_mlperf_training",
```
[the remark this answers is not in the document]
mkuznet1: False positive: madengine treats `built_models.dockerfile` as a prefix and globs `<value>.*` (`docker_builder.py:587`), matching the root `models.json` convention (`"dockerfile": "docker/primus"`). The full path is on line 6, in `built_images.dockerfile`. Verified by a real build + run of both images.

.claude/skills/mad-slurm-multinode/assets/manifests/mlperf_inference_llama-3.1-8b.template.json
```diff
@@ -0,0 +1,91 @@
...
+      "registry_image": null,
+      "registry": null,
+      "gpu_vendor": "AMD"
+    }
+  },
+  "built_models": {
+    "rocm-mlperf-inference-llama31-8b": {
+      "name": "pyt_mlperf_inference_llama-3.1-8b",
+      "url": "",
+      "dockerfile": "docker/pyt_mlperf_inference",
```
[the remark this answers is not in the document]
mkuznet1: False positive: madengine treats `built_models.dockerfile` as a prefix and globs `<value>.*` (`docker_builder.py:587`), matching the root `models.json` convention (`"dockerfile": "docker/primus"`). The full path is on line 6, in `built_images.dockerfile`. Verified by a real build + run of both images.

.claude/skills/mad-slurm-multinode/assets/manifests/mlperf_inference_llama-3.1-8b_smoke.template.json
```diff
@@ -0,0 +1,96 @@
...
+      "registry_image": null,
+      "registry": null,
+      "gpu_vendor": "AMD"
+    }
+  },
+  "built_models": {
+    "rocm-mlperf-inference-llama31-8b": {
+      "name": "pyt_mlperf_inference_llama-3.1-8b",
+      "url": "",
+      "dockerfile": "docker/pyt_mlperf_inference",
```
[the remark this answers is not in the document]
mkuznet1: False positive: madengine treats `built_models.dockerfile` as a prefix and globs `<value>.*` (`docker_builder.py:587`), matching the root `models.json` convention (`"dockerfile": "docker/primus"`). The full path is on line 6, in `built_images.dockerfile`. Verified by a real build + run of both images.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
mkuznet1: Valid, and deliberately not changed here: moving `core_r0.15.0_rocm` to a SHA needs a full image rebuild plus a re-verify run to be honest about what was tested. Tracking it as a follow-up rather than flipping the pin untested.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
mkuznet1: Same answer as `MEGATRON_COMMIT`: valid, but pinning `release_v2.17_rocm` to a SHA is only meaningful after a rebuild + re-verify, so it is a follow-up rather than an untested flip.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
mkuznet1: Agreed, and it is a feature gap rather than a defect: the sglang overlay has `ARG RCCL_*` (`sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile:62-65`) and the training image should too. Out of scope for the MLPerf bring-up; better as its own change on `mad-rccl`.

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
mkuznet1: Checked and kept as an assignment, with the reasoning now in the file (`:97`): the ROCm base image leaves `PYTHONPATH` unset, so appending would only add a trailing colon, which puts the run directory on `sys.path`. d148cff

docker/pyt_mlperf_training.ubuntu.amd.Dockerfile
mkuznet1: Fixed: `|| true` dropped, so a missing `libbnxt_re-rdmav*.so` now fails the build (`docker/pyt_mlperf_training.ubuntu.amd.Dockerfile:383`). d148cff, and the image was rebuilt with the gate in place.

docker/pyt_mlperf_inference.ubuntu.amd.Dockerfile
mkuznet1: Fixed: `evaluate` is installed explicitly at `:59`, with a comment that the base image only happens to ship it. d148cff, verified by a full image rebuild.

docker/pyt_mlperf_inference.ubuntu.amd.Dockerfile
mkuznet1: Fixed: the return value is checked now — `missing = [p for p in ('punkt','punkt_tab') if not nltk.download(...)]` then `sys.exit(...)`, `:68`. d148cff, and the rebuild exercised it.

scripts/pyt_mlperf_training/run.sh
mkuznet1: Fixed: the whole token window (assignment, `-z` test, download) is wrapped in `{ set +x; } 2>/dev/null` with `set -x` restored on both exits (`run.sh:157`, 9723a14). Reproduced the leak first, then confirmed 0 token hits in the trace afterwards.

scripts/pyt_mlperf_training/run.sh
mkuznet1: Fixed: the downloader goes to a temp file with a `-s` size check, and both sentinels are gated on `_tokenizer_present`/`_dataset_present` (`run.sh:196`, 9723a14). Repro: a 404 now returns 1 before the sentinel instead of exit 0 after it.

scripts/pyt_mlperf_training/extract_perf.py
mkuznet1: Fixed: the ternary is collapsed to the constant and a comment says the marker reports launch, not convergence (`extract_perf.py:496`). The run verdict is a separate row — the 2-node run emits `run_stop_aborted`. d148cff

scripts/pyt_mlperf_training/extract_perf.py
mkuznet1: Fixed: `peak_bf16_tflops` is written next to `mfu_pct` (`extract_perf.py:433`, `:459`). The 2-node run gives `peak_bf16_tflops,2500`, `mfu_pct,44.8361` and `per_gpu_tflops_bf16,1120.9` — 1120.9/2500 checks out, so the divisor is now auditable. d148cff

.claude/skills/mad-slurm-multinode/scripts/validate_manifest.sh
mkuznet1: Fixed: `nodelist_min_count()` is a third node-count source, and a bracket range returns 2 so it never understates (`validate_manifest.sh:67`). Verified a nodelist-only 2-node manifest now FAILs on the missing HCA while single-node stays relaxed. 9723a14

.claude/skills/mad-slurm-multinode/assets/manifests/mlperf_training_llama-3.1-8b_smoke.template.json
mkuznet1: Fixed: the placeholder RDMA block is gone from both env blocks of the smoke template — a single node uses xGMI/P2P and never reads those vars. `manifests.md` now says what to copy back when scaling past one node. 9723a14

.claude/skills/mad-slurm-multinode/references/manifests.md
```diff
@@ -173,9 +173,58 @@ differ, the host path belongs on the value side.
...
 Change `slurm.nodes`, `distributed.nnodes`, and the `nodelist` cardinality
 together. Everything else (env, mounts, image) stays the same. Real examples
 exist at 2 and 4 nodes for primus and sglang_disagg.
+
+Scaling a *single-node* template past one node is the one case that needs more
+than the node count: the RDMA transport block is absent by design, so copy
+`NCCL_IB_HCA`, `NCCL_IB_GID_INDEX`, `NCCL_NET`, `NCCL_IB_DISABLE`,
+`RDMAV_DRIVERS`, `IBV_DRIVERS` (and `RCCL_AINIC_ROCE` where the archetype needs
+it) from the multi-node template of the same workload into **both** env blocks.
+`validate_manifest.sh` enforces this: it only waives the transport vars while
```
i-kosarev: It only enforces `NCCL_IB_HCA` — the other six are unchecked (`NCCL_NET`/`NCCL_IB_DISABLE`/`NCCL_IB_GID_INDEX` don't appear in the script at all, and the AINIC trio is symmetry-only, so absent-from-both passes). A 2-node manifest with just `NCCL_IB_HCA` copied validates `0 FAIL`. We might silently get wrong numbers

i-kosarev (commented):
one follow-up after re-reviewing the fixes. everything else checks out!

.claude/skills/mad-slurm-multinode/references/manifests.md
mkuznet1: Confirmed, a 2-node manifest with only NCCL_IB_HCA did validate 0 FAIL. bc803bb: symmetry now also covers NCCL_NET / NCCL_IB_DISABLE / NCCL_IB_GID_INDEX, FAIL on an HCA list next to NCCL_IB_DISABLE=1 or NCCL_NET=Socket, WARN when HCA is the only transport var present. Absent stays a WARN since the needed set is archetype-specific. The doc no longer claims enforcement. All 9 templates unchanged.

i-kosarev (approved):
LGTM

PR #219, opened 2026-08-21
Add vLLM + DeepEP expert-parallel inference Dockerfile

i-kosarev:
## Summary

Adds `docker/vllm_deepep_v2_inference.ubuntu.amd.Dockerfile`: a vLLM image that
serves mixture-of-experts models over DeepEP v2 on MI350X (gfx950). Validated
with DeepSeek-R1 FP8 at `TP=1 / DP=8 / EP=8` on a single node, and benchmarked
on DeepSeek-V2-Lite.

Follows the staging pattern of `sglang_disagg_inference_full_overlay`: every
component is a gated stage, so a rebuild only pays for what changed.

- `DEEPEP_REPO` / `DEEPEP_COMMIT` — DeepEP. Required, and taken as build args
  rather than vendored, since the source is not public.
- `AITER_COMMIT` — AITER plus the zero-token `get_ksplit` guard, which a
  data-parallel profile microbatch can otherwise trip into a division by zero.
- `RDMA_CORE_VERSION` — rdma-core from source, default `63.0`, replacing the
  distro packages and dropping the vendor `bnxt_re` provider. An empty value
  keeps the base image's.
- `ENABLE_TORCHVISION_STUB` — text-only torchvision surface, for ROCm bases
  whose torchvision ABI breaks `import vllm` through a multimodal config.

## Runtime notes captured in the header

These are the settings that cost the most time to rediscover:

- `NCCL_CUMEM_ENABLE=1` is mandatory. Without VMM, RCCL answers
  `Communicator does not support symmetric memory` and the ElasticBuffer is
  never created.
- `EP_DISABLE_GIN=1` on a single node. DeepEP v2 moves no payload through GIN
  inside one XGMI domain — `is_scaleup_nvlink` is true and the barrier and
  dispatch/combine run over symmetric memory — but the buffer constructor still
  asserts on `ginType` and reserves queue pairs unless GIN is disabled. Leaving
  it on additionally requires `/dev/infiniband`, `NCCL_GIN_TYPE=2` and
  `EP_GIN_QUEUE_DEPTH=0`.
- Mount the JIT caches. AITER tunes GEMM shapes at run time behind a global
  build lock; with data parallelism one rank tunes while the rest wait in the
  collective, which shows up as P99 TTFT in the tens of seconds until the cache
  covers the shapes in play.

## Test

Built and run on 8x MI350X. DeepSeek-R1 FP8 generates correctly. DeepSeek-V2-Lite
reaches 633 / 1238 / 2371 total tok/s at concurrency 8 / 16 / 32 for
`in=256, out=128`, measured after the tuning cache converges.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docker/vllm_deepep_v2_inference.ubuntu.amd.Dockerfile
+++ b/docker/vllm_deepep_v2_inference.ubuntu.amd.Dockerfile
@@ -1,0 +1,217 @@
+# CONTEXT {'gpu_vendor': 'AMD', 'guest_os': 'UBUNTU'}
+###############################################################################
+#
+# MIT License
+#
+# Copyright (c) 2026 Advanced Micro Devices, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


i-kosarev:
Superseded by #220 (same change, cleaner branch).

PR #226, opened 2026-08-24
Add KV_OFFLOAD overlay for vllm disagg P/D with NIXL connector

nitsingh2:
Adds an orthogonal `KV_OFFLOAD` axis that layers a KV tier on top of the disagg P/D connector via vLLM's `MultiConnector`. 
 The base P/D transport stays **NIXL**; the offload sub-connector is listed first, so a worker checks its local tier before the P→D NIXL fetch. `KV_OFFLOAD=none` is a no-op. Validated on **Llama-3.1-405B-Instruct-FP8-KV** (1P1D, TP=8).

  ## Architecture — two independent axes
  - **Base connector (node axis, P→D):** NIXL — unchanged.
  - **Offload overlay (tier axis, within a node):** `OFFLOAD_BACKEND` × tier.

  ## Matrix
  The code functionally supports this full matrix; but only `lmcache` + `cpu` is currently in optimal state.

  | Offloading Connector | Tier | KV transfer conenctor| optimal |
  |---------|------|-----------|---------------------|
  | `native` | CPU | RIXL | no |
  | `native` | CPU + FS | RIXL | no |
  | `lmcache`  | CPU | RIXL | yes |
  | `lmcache` | CPU + FS | RIXL | no |


scripts/vllm_dissag/connectors/moriio.sh
```diff
@@ -231,7 +231,7 @@ connector_launch_worker() {
                     "${mem_args[@]}" \
                     --kv-cache-dtype "${_kvdtype}" \
                     --block-size "${_block}" \
-                    ${_prefix_cache_flag} \
+                    --no-enable-prefix-caching \
```
lcskrishna: Try to add it as an option to user - whether to enabel Prefix caching or not. I believe there is one. Double check this line.

scripts/vllm_dissag/connectors/moriio.sh
```diff
@@ -251,7 +251,7 @@ connector_launch_worker() {
             "${mem_args[@]}" \
             --kv-cache-dtype "${_kvdtype}" \
             --block-size "${_block}" \
-            ${_prefix_cache_flag} \
+            --no-enable-prefix-caching \
```
lcskrishna: same as above.

lcskrishna (changes requested):
Few comments and like discussed run with a different model.

